PR #3, opened 2022-01-19
Copy fundamental functionality from `github.com/elastic/beats/v7/libbeat/common`

kvch:
I choose `logp` and `logp/configure` pacakges to be moved first because it barely has any dependencies in libbeat, but almost every package imports it.

### `logp` dependencies
logp
One of the packages I need before `logp` can be moved is `common.Config`.

### `logp/configure` dependencies
configure

Although, Elastic Agent does not use directly `common.Config` in many places, packages it uses require it. The other dependency from `libbeat` in the graphs above is `github.com/elastic/beats/v7/libbeat/paths` that includes small helpers to determine the paths to logs, data and config files. Those dependencies are going to be transformed and passed as parameters instead of importing them.

The following packages are copied in this PR:

* `common.Config` from libbeat to `config.C`
* `StringSet` from libbeat to `str.Set`

I renamed the structs to be more idiomatic.


kvch:
jenkins run tests

README.md
```diff
@@ -1,2 +1,7 @@
 # elastic-agent-libs
-Libraries used by Elastic Agent and Beats
+
+This repository is the home to the common libraries used by Elastic Agent and Beats.
+
+Provided packages:
+* `github.com/elastic/elastic-agent-libs/config` the previous `config.go` file from `github.com/elastic/beats/v7/libbeat/common`. A minimal wrpper aroung `github.com/elastic/go-ucfg`. It contains helpers for merging and accessing configuration objects.
```
rdner, line 6:
```suggestion
* `github.com/elastic/elastic-agent-libs/config` the previous `config.go` file from `github.com/elastic/beats/v7/libbeat/common`. A minimal wrapper around `github.com/elastic/go-ucfg`. It contains helpers for merging and accessing configuration objects.
```

config/config.go
```diff
@@ -0,0 +1,377 @@
...
+// MustNewConfigFrom panics if an error occurs.
+func MustNewConfigFrom(from interface{}) *C {
+	cfg, err := NewConfigFrom(from)
+	if err != nil {
+		panic(err)
+	}
+	return cfg
+}
+
+func MergeConfigs(cfgs ...*C) (*C, error) {
```
rdner, line 104: Should we add comments to the rest of the exported functions?

rdner (changes requested):
Some typos to fix and perhaps more GoDoc (no strong opinion about the doc).

rdner (approved):

PR #10, opened 2022-02-16
Create linter configuration and address found issues

rdner:
This PR is to propose the linting configuration that we are going to re-use for new repositories.

Also, all the currently found linting issues are fixed in this PR which can give an idea what kind of common issues we have now in our code and which the linter finds.

Closes https://github.com/elastic/beats/issues/30150

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -37,11 +37,6 @@ pipeline {
         withGithubNotify(context: "Lint") {
           dir("${BASE_DIR}"){
             withMageEnv(){
-              sh(label: 'lint', script: '''
-                go mod tidy && git diff --exit-code
-								curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.0
-                ./bin/golangci-lint run ./...
-              ''')
               sh(label: 'Mage check', script: 'mage check')
             }
           }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -39,10 +39,10 @@ pipeline {
             withMageEnv(){
               sh(label: 'lint', script: '''
                 go mod tidy && git diff --exit-code
-                gofmt -l . | read && echo "Code differs from gofmt's style. Run 'gofmt -w .'" 1>&2 && exit 1 || true
+								curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.0
+                ./bin/golangci-lint run ./...
               ''')
               sh(label: 'Mage check', script: 'mage check')
-              sh(label: 'Go vet', script: 'go vet ./...')
             }
           }
         }
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -47,8 +47,6 @@ linters:
     - importas # enforces consistent import aliases
     - nilerr # finds the code that returns nil even if it checks that the error is not nil.
     - noctx # noctx finds sending http request without context.Context
-    - rowserrcheck # checks whether Err of rows is checked successfully
-    - sqlclosecheck # checks that sql.Rows and sql.Stmt are closed.
     - unconvert # Remove unnecessary type conversions
     - wastedassign # wastedassign finds wasted assignment statements.
     - godox # tool for detection of FIXME, TODO and other comment keywords
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,153 @@
+# options for analysis running
+run:
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 1m
+
+output:
+  sort-results: true
+
+# Uncomment and add a path if needed to exclude
+# skip-dirs:
+#   - some/path
+# skip-files:
+#   - ".*\\.my\\.go$"
+#   - lib/bad.go
+
+# Find the whole list here https://golangci-lint.run/usage/linters/
+linters:
+  disable-all: true
+  enable:
+    - deadcode # finds unused code
+    - errcheck # checking for unchecked errors in go programs
+    - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
+    # - goconst # finds repeated strings that could be replaced by a constant
+    - dupl # tool for code clone detection
+    - forbidigo # forbids identifiers	by reg exps
+    - gomoddirectives # manage the use of 'replace', 'retract', and 'excludes' directives in go.mod.
+    - gosimple # linter for Go source code that specializes in simplifying a code
+    - misspell # finds commonly misspelled English words in comments
+    - nakedret # finds naked returns in functions greater than a specified function length
+    - prealloc # finds slice declarations that could potentially be preallocated
+    - nolintlint # reports ill-formed or insufficient nolint directives
+    - staticcheck # Staticcheck is a go vet on steroids, applying a ton of static analysis checks
+    - stylecheck # a replacement for golint
+    - unparam # reports unused function parameters
+    - unused # checks Go code for unused constants, variables, functions and types
+
+    - govet # Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
+    - ineffassign # detects when assignments to existing variables are not used
+    - structcheck # finds unused struct fields
+    - typecheck # Like the front-end of a Go compiler, parses and type-checks Go code
+    - varcheck # Finds unused global variables and constants
+    - asciicheck # simple linter to check that your code does not contain non-ASCII identifiers
+    - bodyclose # checks whether HTTP response body is closed successfully
+    - durationcheck # check for two durations multiplied together
+    - exportloopref # checks for pointers to enclosing loop variables
+    - forcetypeassert # finds forced type assertions
+    - goimports # Goimports does everything that gofmt does. Additionally it checks unused imports
+    - gosec # inspects source code for security problems
+    - importas # enforces consistent import aliases
+    - nilerr # finds the code that returns nil even if it checks that the error is not nil.
+    - noctx # noctx finds sending http request without context.Context
+    - rowserrcheck # checks whether Err of rows is checked successfully
+    - sqlclosecheck # checks that sql.Rows and sql.Stmt are closed.
+    - unconvert # Remove unnecessary type conversions
+    - wastedassign # wastedassign finds wasted assignment statements.
+    - godox # tool for detection of FIXME, TODO and other comment keywords
+
+# all available settings of specific linters
+linters-settings:
+  errcheck:
+    # report about not checking of errors in type assertions: `a := b.(MyStruct)`;
+    # default is false: such cases aren't reported by default.
+    check-type-assertions: true
+
+  errorlint:
+    # Check whether fmt.Errorf uses the %w verb for formatting errors. See the readme for caveats
+    errorf: true
+    # Check for plain type assertions and type switches
+    asserts: true
+    # Check for plain error comparisons
+    comparison: true
+
+  goconst:
+    # minimal length of string constant, 3 by default
+    min-len: 3
+    # minimal occurrences count to trigger, 3 by default
+    min-occurrences: 2
+
+  makezero:
+    # Allow only slices initialized with a length of zero. Default is false.
+    always: true
+
+  dupl:
+    # tokens count to trigger issue, 150 by default
+    threshold: 100
+
+  forbidigo:
+    # Forbid the following identifiers
+    forbid:
+      - fmt.Print.* # too much log noise
+    # Exclude godoc examples from forbidigo checks.  Default is true.
+    exclude_godoc_examples: true
+
+  gomoddirectives:
+    # Allow local `replace` directives. Default is false.
+    replace-local: false
+
+  gosimple:
+    # Select the Go version to target. The default is '1.13'.
+    go: "1.17"
+
+  misspell:
+    # Correct spellings using locale preferences for US or UK.
+    # Default is to use a neutral variety of English.
+    # Setting locale to US will correct the British spelling of 'colour' to 'color'.
+    locale: US
+    # ignore-words:
+    #   - IdP
+
+  nakedret:
+    # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
+    max-func-lines: 0
+
+  prealloc:
+    # XXX: we don't recommend using this linter before doing performance profiling.
+    # For most programs usage of prealloc will be a premature optimization.
+
+    # Report preallocation suggestions only on simple loops that have no returns/breaks/continues/gotos in them.
+    # True by default.
+    simple: true
+    range-loops: true # Report preallocation suggestions on range loops, true by default
+    for-loops: false # Report preallocation suggestions on for loops, false by default
+
+  nolintlint:
+    # Enable to ensure that nolint directives are all used. Default is true.
+    allow-unused: false
+    # Disable to ensure that nolint directives don't have a leading space. Default is true.
+    allow-leading-space: true
+    # Exclude following linters from requiring an explanation.  Default is [].
+    allow-no-explanation: []
+    # Enable to require an explanation of nonzero length after each nolint directive. Default is false.
+    require-explanation: true
+    # Enable to require nolint directives to mention the specific linter being suppressed. Default is false.
+    require-specific: true
+
+  staticcheck:
+    # Select the Go version to target. The default is '1.13'.
+    go: "1.17"
+
+  stylecheck:
+    # Select the Go version to target. The default is '1.13'.
+    go: "1.17"
+
+  unparam:
+    # Inspect exported functions, default is false. Set to true if no external program/library imports your code.
+    # XXX: if you enable this setting, unparam will report a lot of false-positives in text editors:
+    # if it's called for subdir of a project it can't find external interfaces. All text editor integrations
+    # with golangci-lint call it on a directory with the changed file.
+    check-exported: false
+
+  unused:
+    # Select the Go version to target. The default is '1.13'.
+    go: "1.17"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,7 @@ linters:
     - deadcode # finds unused code
     - errcheck # checking for unchecked errors in go programs
     - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
+    - goconst # finds repeated strings that could be replaced by a constant
     - dupl # tool for code clone detection
     - forbidigo # forbids identifiers	by reg exps
     - gomoddirectives # manage the use of 'replace', 'retract', and 'excludes' directives in go.mod.
@@ -65,6 +66,12 @@ linters-settings:
     asserts: true
     # Check for plain error comparisons
     comparison: true
+
+  goconst:
+    # minimal length of string constant, 3 by default
+    min-len: 3
+    # minimal occurrences count to trigger, 3 by default
+    min-occurrences: 2
 
   dupl:
     # tokens count to trigger issue, 150 by default
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,7 +22,7 @@ linters:
     - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
     - goconst # finds repeated strings that could be replaced by a constant
     - dupl # tool for code clone detection
-    - forbidigo # forbids identifiers	by reg exps
+    - forbidigo # forbids identifiers	matched by reg exps
     - gomoddirectives # manage the use of 'replace', 'retract', and 'excludes' directives in go.mod.
     - gosimple # linter for Go source code that specializes in simplifying a code
     - misspell # finds commonly misspelled English words in comments
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,7 +20,6 @@ linters:
     - deadcode # finds unused code
     - errcheck # checking for unchecked errors in go programs
     - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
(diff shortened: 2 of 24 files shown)


logp/config.go
```diff
@@ -41,7 +41,6 @@ type Config struct {
 	environment Environment
 	addCaller   bool // Adds package and line number info to messages.
 	development bool // Controls how DPanic behaves.
-	logsPath    string
```
rdner, line 44: This was unused.

logp/core.go
```diff
@@ -246,13 +246,13 @@ func makeFileOutput(cfg Config) (zapcore.Core, error) {
...
-	return newCore(cfg, buildEncoder(cfg), rotator, cfg.Level.ZapLevel()), nil
+	return newCore(buildEncoder(cfg), rotator, cfg.Level.ZapLevel()), nil
 }
 
-func newCore(cfg Config, enc zapcore.Encoder, ws zapcore.WriteSyncer, enab zapcore.LevelEnabler) zapcore.Core {
-	return wrappedCore(cfg, zapcore.NewCore(enc, ws, enab))
+func newCore(enc zapcore.Encoder, ws zapcore.WriteSyncer, enab zapcore.LevelEnabler) zapcore.Core {
+	return wrappedCore(zapcore.NewCore(enc, ws, enab))
 }
-func wrappedCore(cfg Config, core zapcore.Core) zapcore.Core {
```
rdner, line 255: the parameter was unused.

.ci/Jenkinsfile
```diff
@@ -39,10 +39,10 @@ pipeline {
             withMageEnv(){
               sh(label: 'lint', script: '''
                 go mod tidy && git diff --exit-code
-                gofmt -l . | read && echo "Code differs from gofmt's style. Run 'gofmt -w .'" 1>&2 && exit 1 || true
+								curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.0
```
kvch, line 42: Given that we are adding more and more linting to this step, it would be better to extract it into a separate mage step. Maybe it can be added to `mage check` and we could remove this step from the CI.

logp/config.go
kvch: Yes, it is a leftover from the migration of the package. Thanks.

.ci/Jenkinsfile
ph: + 1 to a separate step.

logp/core.go
ph: This will impact beats/fleet I think

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+    # ignore-words:
+    #   - IdP
+
+  nakedret:
+    # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
+    max-func-lines: 0
+
+  prealloc:
+    # XXX: we don't recommend using this linter before doing performance profiling.
+    # For most programs usage of prealloc will be a premature optimization.
```
ph, line 116: @rdner Looking at the comment we might want to remove it?

.ci/Jenkinsfile
rdner: @ph @kvch 
but does it really make any difference to have the linter as a separate step in Jenkins (as it is now) or as a separate step in the Magefile?

Mind you, I have not added more linting to the existing step, I replaced the linter we already had with a new one (because it includes the same checks, so we don't need the old one anymore).

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+    # Disable to ensure that nolint directives don't have a leading space. Default is true.
+    allow-leading-space: true
+    # Exclude following linters from requiring an explanation.  Default is [].
+    allow-no-explanation: []
+    # Enable to require an explanation of nonzero length after each nolint directive. Default is false.
+    require-explanation: true
+    # Enable to require nolint directives to mention the specific linter being suppressed. Default is false.
+    require-specific: true
+
+  staticcheck:
```
ph, line 136: NIT: It would be nice to have the `1.17` to be populated by the `.go-version` file.

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+#   - lib/bad.go
+
+# Find the whole list here https://golangci-lint.run/usage/linters/
+linters:
+  disable-all: true
+  enable:
+    - deadcode # finds unused code
+    - errcheck # checking for unchecked errors in go programs
+    - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
+    # - goconst # finds repeated strings that could be replaced by a constant
```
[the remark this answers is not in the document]
rdner: even in this small repo it complains a lot about each string value repeated twice. I don't think it's worth it. I looked at some code in beats and it's very common, I don't think we should enforce creating constants for everything but if the team votes differently I'll enable it, hence it's still commented in the config.

.golangci.yml
rdner: Every comment was just copied from the official doc. And it makes sense to me. By default, we should pre-allocate but it does not mean that it's required.

rdner: https://golangci-lint.run/usage/linters/#prealloc

belimawr: I'm not a huge fan of this one either, I'm not a huge fan of premature optimisations. I'm just wondering if it can get noisy.

.golangci.yml
belimawr: In the long run it would be nice to have it reading from `.go-version` so we don't need to remember to update this configuration file as we change our Go version.

Because we're most likely running it with mage, I see no issue on adding this automation.

file/helper_other.go
```diff
@@ -37,8 +37,10 @@ func SafeFileRotate(path, tempfile string) error {
 	// filesystems, so to update the parents directory metadata to actually
 	// contain the new file being rotated in.
 	f, err := os.Open(parent)
+
+	// nolint: nilerr // ignore error
 	if err != nil {
-		return nil // ignore error
+		return nil
```
belimawr, line 43: I'm not an expert here, but this looks like the kind of error that would be nice to at least log and move on. I can imagine it generating a very hard to debug edge-case bug.

file/rotator.go
```diff
@@ -258,9 +258,9 @@ func (r *Rotator) appendToFile() error {
 		return errors.Wrap(err, "failed to append to existing file")
 	}
 	if r.redirectStderr {
-		RedirectStandardError(r.file)
+		err = RedirectStandardError(r.file)
 	}
-	return nil
+	return err
 }
```
belimawr, line 264: A couple of things:
1. That changes the semantics because now the error from `RedirectStandardError` can be returned. I'm not sure whether that's desired or not.
2. I suggest an early return or to handle the error inside the `if` statement so the last return is clearly a no-error return.

```go
	if r.redirectStderr {
		return RedirectStandardError(r.file)
	}
	return nil
}
```

logp/core.go
belimawr: Is `elastic-agent-libs` intended to be a drop-in replacement for beats/libbeat?

logp/core.go
```diff
@@ -267,7 +267,8 @@ func loadLogger() *coreLogger {
 
 func storeLogger(l *coreLogger) {
 	if old := loadLogger(); old != nil {
-		old.rootLogger.Sync()
+		// ignore the error
+		_ = old.rootLogger.Sync()
```
belimawr, line 271: That's the reason I'm not a huge fan of this `errcheck`lint... But it's not a strong opinion.

Anyways I believe the `// ignore the error` comment is redundant here, the error is clearly being ignored.

logp/core_test.go
```diff
@@ -46,9 +46,11 @@ func TestLogger(t *testing.T) {
...
 	}
 
-	TestingSetup()
+	err := TestingSetup()
+	assert.NoError(t, err)
 	exerciseLogger()
-	TestingSetup()
+	err = TestingSetup()
+	assert.NoError(t, err)
 	exerciseLogger()
```
belimawr, line 54: I believe this should be `require` instead of `assert`. Does it make sense to continue the test if `TestingSetup` fails?

logp/global_test.go
```diff
@@ -80,7 +80,8 @@ func TestRecover(t *testing.T) {
 	const recoveryExplanation = "Something went wrong"
 	const cause = "unexpected condition"
 
-	DevelopmentSetup(ToObserverOutput())
+	err := DevelopmentSetup(ToObserverOutput())
+	assert.NoError(t, err)
```
belimawr, line 84: Same thing here (`require` instead of `assert`), does it make sense to continue the test if the development setup fails?
```suggestion
	require.NoError(t, err)
```

logp/selective_test.go
```diff
@@ -24,7 +24,8 @@ import (
 )
 
 func TestHasSelector(t *testing.T) {
-	DevelopmentSetup(WithSelectors("*", "config"))
+	err := DevelopmentSetup(WithSelectors("*", "config"))
+	assert.NoError(t, err)
```
belimawr, line 28: Same thing, should we use `require`?

paths/paths.go
```diff
@@ -87,7 +89,7 @@ func (paths *Path) InitPaths(cfg *Path) error {
 	// make sure the data path exists
 	err = os.MkdirAll(paths.Data, 0750)
 	if err != nil {
-		return fmt.Errorf("Failed to create data path %s: %v", paths.Data, err)
+		return errors.Wrapf(err, "Failed to create data path %s", paths.Data)
```
belimawr, line 92: Question: Why not use `%w` instead of `github.com/pkg/errors`?

belimawr (commented):
Great job @rdner !!

One little detail, once the PR is finalised, what about sorting the liners alphabetically? I feel it makes reading easier.

logp/core.go
rdner: @ph how come? this is a private function and the entire package is in this repo now, is not it?

.ci/Jenkinsfile
rdner: Good point, I'll move it to the magefile.

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+    - ineffassign # detects when assignments to existing variables are not used
+    - structcheck # finds unused struct fields
+    - typecheck # Like the front-end of a Go compiler, parses and type-checks Go code
+    - varcheck # Finds unused global variables and constants
+    - asciicheck # simple linter to check that your code does not contain non-ASCII identifiers
+    - bodyclose # checks whether HTTP response body is closed successfully
+    - durationcheck # check for two durations multiplied together
+    - exportloopref # checks for pointers to enclosing loop variables
+    - forcetypeassert # finds forced type assertions
+    - goimports # Goimports does everything that gofmt does. Additionally it checks unused imports
```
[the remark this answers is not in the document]
rdner: @belimawr what you're suggesting is more for an editor, is not it?

This just checks that the imports are used and tells you if there is something wrong, it does not change your code. Although the linter has the `fix: true` configuration parameters, I think we should fix all the linting issues manually to understand them better and to avoid repeating.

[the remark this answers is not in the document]
rdner: I'm just concerned that using `gci` as a linter would force every single member of our team to run it in their editor, otherwise it would not satisfy the check.

.golangci.yml
rdner: I believe if it's known one should always set capacity when allocating, for example:

```go
b := map[string]interface{}{}
a := make([]string, 0, len(b)) // vs `var a []string`

for key := range b {
  a = append(a, key)
}
```

file/helper_other.go
rdner: It's kind of out of scope of this PR, I think :)

logp/core.go
ph: I've missed the private function part.

.golangci.yml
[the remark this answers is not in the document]
ph: Well, this is still a code smell and prevent typos error.

.golangci.yml
ph: @rdner This shouldn't be a lot of effort,  we have already code generation. There is a `make update` task that generates the configuration files from templates and other generated files. That `make update` comment is run on every build, if `git status` detects new files or modified files it will fail the build.

So the workflow would be like this.

- Update `.go-version`
- run `make update`
- Push PR with the updated linter information and the updated go version.
- We have an automated task that does this https://github.com/elastic/fleet-server/pull/1143

But on the other side we will have 1.17 for some time and 1.18 for quite some time, I just hate having multiple files to updates we still have too much and removing the human in the operation is always a good thing and having multiple go repository that just work the same would save us time.

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+    - unconvert # Remove unnecessary type conversions
+    - wastedassign # wastedassign finds wasted assignment statements.
+    - godox # tool for detection of FIXME, TODO and other comment keywords
+
+# all available settings of specific linters
+linters-settings:
+  errcheck:
+    # report about not checking of errors in type assertions: `a := b.(MyStruct)`;
+    # default is false: such cases aren't reported by default.
+    check-type-assertions: true
```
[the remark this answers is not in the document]
rdner: You right, they do the same check. I removed `forcetypeassert`.

file/rotator.go
rdner: I really thought it was a bug but you're right, I should not have changed the behaviour. I rolled the change back and use the early return you suggested.

.golangci.yml
rdner: I removed the comment that it would scare anyone. I'd like to keep the linter though unless everyone are oppose it.

.golangci.yml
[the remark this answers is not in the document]
rdner: I'd like to see more opinions from the rest of the team, I don't have a strong opinion against adding this linter.

.golangci.yml
```diff
@@ -0,0 +1,138 @@
...
+
+# Find the whole list here https://golangci-lint.run/usage/linters/
+linters:
+  disable-all: true
+  enable:
+    - deadcode # finds unused code
+    - errcheck # checking for unchecked errors in go programs
+    - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
+    - dupl # tool for code clone detection
+    - forbidigo # forbids identifiers	by reg exps
```
cmacknz, line 24:
```suggestion
    - forbidigo # forbids identifiers used by reg exps
```

.golangci.yml
```diff
@@ -0,0 +1,138 @@
...
+    - asciicheck # simple linter to check that your code does not contain non-ASCII identifiers
+    - bodyclose # checks whether HTTP response body is closed successfully
+    - durationcheck # check for two durations multiplied together
+    - exportloopref # checks for pointers to enclosing loop variables
+    - goimports # Goimports does everything that gofmt does. Additionally it checks unused imports
+    - gosec # inspects source code for security problems
+    - importas # enforces consistent import aliases
+    - nilerr # finds the code that returns nil even if it checks that the error is not nil.
+    - noctx # noctx finds sending http request without context.Context
+    - rowserrcheck # checks whether Err of rows is checked successfully
```
cmacknz, line 50: Is it worth having this and `sqlclosecheck` running in repositories that don't use the sql driver at all?

paths/paths.go
cmacknz: `github.com/pkg/errors` is archived considers itself replaced by the standard library functionality. https://github.com/pkg/errors is archived for example.

cmacknz:
Thanks for proposing this @rdner. There is a lot to deal with to get this enabled it seems.

paths/paths.go
rdner: Interesting, I missed that. `errors.Wrap` is quite common in the codebase, I'll migrate the rest of the code to `fmt.Errorf` then.

rdner: I replaced all 32 occurrences in the entire repo (97bd94a3098995057bdd1ec6d7b250597152703c) and `github.com/pkg/errors` is no longer a direct dependency.

.golangci.yml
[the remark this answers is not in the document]
rdner: @belimawr I tried to run the code base against `gci` and I get `File is not `gci`-ed (gci)` for most of the files where `goimports` passes just fine. I'm not in favour, I don't think grouping imports is so important.

.golangci.yml
rdner: @ph This PR already has a ton of comments, I'm not going to implement it here but I can take as the next step.

.golangci.yml
rdner: Removed.

.golangci.yml
rdner: I disagree with this, `forbidigo` is using regular expressions to forbid certain identifiers. It's rather `forbids identifiers matched by reg exps` but I think in its initial form it's clear enough.

.ci/Jenkinsfile
rdner: Addressed in af131c82958bf60e3f1c1a5afd480b5807090b5b

.golangci.yml
rdner: Fixed in bc2c107eb5ca81333cf5f5d549143aa7f7e8987f

.ci/Jenkinsfile
```diff
@@ -37,12 +37,7 @@ pipeline {
         withGithubNotify(context: "Lint") {
           dir("${BASE_DIR}"){
             withMageEnv(){
-              sh(label: 'lint', script: '''
-                go mod tidy && git diff --exit-code
-                gofmt -l . | read && echo "Code differs from gofmt's style. Run 'gofmt -w .'" 1>&2 && exit 1 || true
-              ''')
```
rdner, line 43: * `go mod tidy` has been moved to the magefile
* `gofmt` is now included into the new linter

.ci/Jenkinsfile
```diff
@@ -37,12 +37,7 @@ pipeline {
...
               sh(label: 'Mage check', script: 'mage check')
-              sh(label: 'Go vet', script: 'go vet ./...')
```
rdner, line 45: Is not included into the new linter.

file/rotator.go
```diff
@@ -380,7 +384,12 @@ func (r *Rotator) closeFile() error {
 	}
 	err := r.file.Close()
 	r.file = nil
-	return errors.Wrap(err, "failed to close active file")
```
rdner, line 383: `Wrap` returns `nil` if the `err == nil`, `fmt.Errorf` does not, so had to adjust.

ph (approved):

cmacknz (approved):

.golangci.yml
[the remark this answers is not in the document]
belimawr: > @belimawr I tried to run the code base against `gci` and I get `File is not `gci`-ed (gci)` for most of the files where `goimports` passes just fine. I'm not in favour, I don't think grouping imports is so important.

I haven't tried it on Beats, in my previous job the projects were small and we were adding imports all the time, so it helped to keep things nice and tidy. Otherwise we'd  end up with our imports split into 20 different groups with no meaning.

> I'm just concerned that using gci as a linter would force every single member of our team to run it in their editor

That's not quite true because it could be part of `mage fmt` which we already have to run, so it'd be mostly transparent.

Anyway, that's far from a must-have to me. If I'm the only one who likes it, I'm happy to go without it ;)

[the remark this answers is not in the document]
cmacknz: For what it is worth, I also like grouping imports, but not enough to force it on people. 

Import grouping isn't the most important problem in the code the linter will find. We could look at enabling this later after we have the linter enabled everywhere (if we find it is annoying).

.golangci.yml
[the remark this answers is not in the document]
belimawr: To be honest, I've seen codebases using both approaches and the ones without the constants were easier to read/understand because one did not have to be jumping back and forth between the constant usage and definitions to know their value/visible result.

@ph does have a point regarding typos. Again, not a strong opinion.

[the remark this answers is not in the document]
cmacknz: I don't have a strong opinion on this, I'm probably fine with either option.

For some of the linters we have that are more cosmetic or controversial, we could consider enabling them in a second pass after we have the linter setup everywhere. 

It is easier to reason about the consequences of enabling a particular linter if we can see what problems it is finding in a diff.

faec (approved):
🎉

belimawr (commented):
I know it has been merged, but I'm still submitting it so I don't lose all I've written :P

magefile.go
```diff
@@ -61,3 +119,17 @@ func CheckNoBeatsDependency() error {
...
+	if err != nil {
+		return err
+	}
+	err = sh.Run("git", "diff", "--exit-code")
+	if err != nil {
+		return fmt.Errorf("`go mod tidy` was not called before committing: %w", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
rdner: It just needs `"go.mod"` as the last argument on line 129, I'll add in a follow-up.

logp/environment.go
```diff
@@ -43,21 +43,30 @@ const (
...
 
+const (
+	defaultEnvironmentString        = "default"
+	systemdEnvironmentString        = "systemd"
+	containerEnvironmentString      = "container"
+	macOSServiceEnvironmentString   = "macOS_service"
+	windowsServiceEnvironmentString = "windows_service"
+	invalidEnvironmentString        = "<invalid>"
+)
+
```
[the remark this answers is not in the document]
rdner: Feel free to open a PR and move the discussion there :)

magefile.go
```diff
@@ -23,15 +23,73 @@ package main
...
+	"github.com/magefile/mage/sh"
 )
 
+const (
+	linterInstallFilename = "./intall-golang-ci.sh"
+	linterBinaryFilename  = "./bin/golangci-lint"
+	linterVersion         = "v1.44.0"
+)
+
+// InstallLinter installs golangci-lint (https://golangci-lint.run) to `./bin`
```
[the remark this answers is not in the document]
rdner: I'll switch to `./build` in my follow-up PR, no problem. Didn't know we used `./build` it's more common to have `./bin`.

PR #13, opened 2022-02-23
Move common.MapStr and rename it to mapstr.M

kvch:
This PR moves the struct `MapStr` from `github.com/elastic/beats/v7/libbeat/common`. I put the type under a new package and renamed it to `mapstr.M` to avoid stuttering. I am open to keeping the original type name `mapstr.MapStr`, I just felt it is more idiomatic. Given that we have to rename it everywhere in beats and agent, it does not really matter how we rename it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/elastic/go-ucfg v0.8.4
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/magefile/mage v1.12.1
+	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.3.0
 	github.com/stretchr/testify v1.7.0
 	go.elastic.co/ecszap v1.0.0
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/elastic/go-ucfg v0.8.4
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/magefile/mage v1.12.1
+	github.com/pkg/errors v0.9.1
```
rdner, line 9: I've been asked to not used this deprecated package anymore, see https://github.com/elastic/elastic-agent-libs/pull/10#discussion_r808796031

rdner (approved):

cmacknz (approved):

ph:
I like the change, but I am not sure about the impact on the actual project. We could add an alias to the old type to ease the transition?

kvch:
> I like the change, but I am not sure about the impact on the actual project. We could add an alias to the old type to ease the transition?

I trust the tools available for refactoring code and renaming variables. The effort for the migration and reviewing the change should be minimal. I would add an alias only if we see that people struggle to adopt the new type name.

ph:
@kvch well I was proposing that to reduce the moving changes, and it doesn't cost that much. It was a suggestion.

PR #15, opened 2022-02-24
Move `atomic` package

kvch:
I turned off a warning for the tests because it made no sense. The error was due to duplicated code, but we cannot reuse the code because it applies to different types.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,12 +13,6 @@ issues:
   # Default: 50
   max-issues-per-linter: 0
 
-  exclude-rules:
-    # tests for types that are only sometimes equivalent
-    # (ex: uint and uint64) result in false positive lint warnings
-    path: atomic/atomic_test.go
-    text: "lines are duplicate of"
-
 output:
   sort-results: true
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,8 +13,9 @@ issues:
   # Default: 50
   max-issues-per-linter: 0
 
-  # exclude invalid errors for tests
   exclude-rules:
+    # tests for types that are only sometimes equivalent
+    # (ex: uint and uint64) result in false positive lint warnings
     path: atomic/atomic_test.go
     text: "lines are duplicate of"
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,6 +13,11 @@ issues:
   # Default: 50
   max-issues-per-linter: 0
 
+  # exclude invalid errors for tests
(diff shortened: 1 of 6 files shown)


.golangci.yml
```diff
@@ -13,6 +13,11 @@ issues:
   # Default: 50
   max-issues-per-linter: 0
 
+  # exclude invalid errors for tests
+  exclude-rules:
+    path: atomic/atomic_test.go
+    text: "lines are duplicate of"
```
rdner, line 19: where did you have this reported? does it makes sense to exclude the entire file or maybe adding a single (a few) exception to the line would be enough?

cmacknz: I ran the linter locally and this is the specific output:

```sh
./build/golangci-lint run --new-from-rev=origin/main --whole-files ./... --print-issued-lines=false
atomic/atomic_test.go:55: 55-185 lines are duplicate of `atomic/atomic_test.go:99-229` (dupl)
atomic/atomic_test.go:99: 99-229 lines are duplicate of `atomic/atomic_test.go:143-273` (dupl)
atomic/atomic_test.go:143: 143-273 lines are duplicate of `atomic/atomic_test.go:55-185` (dupl)
atomic/atomic_test.go:187: 187-273 lines are duplicate of `atomic/atomic_test.go:55-141` (dupl)
atomic/atomic_test.go:231: 231-273 lines are duplicate of `atomic/atomic_test.go:55-97` (dupl)
```

It considers large sections of the file to be duplicating other sections multiple times. I think at minimum it is considering the `uint` tests to be functionally the same as the `uint64` tests, which is probably true but that is an implementation detail.

This whole file is like a pathological case for this linter.

cmacknz (approved):

.golangci.yml
cmacknz: I would be in favor of a more specific comment, like "tests for types that are only sometimes equivalent (ex: uint and uint64) result in false positive lint warnings"

rdner: We should just leave an exception comment in the file itself instead of editing the linter configuration file.

Because even in this PR the rendered template has been changed instead of the template itself, the next `.go-version` change will break it. I think we should employ the practice to declare all the exceptions in place with a valid comment. You can do it with:

```go
//nolint:dupl // this is valid for tests
package pkg
```

More details on exceptions can be found here https://golangci-lint.run/usage/false-positives/

rdner (approved):
LGTM now.

PR #19, opened 2022-03-02
Move `github.com/elasitc/beats/v7/common/transport/tlscommon`

kvch:
This PR moves the package mentioned in the title. It has no significant changes compared to what we have in libbeat.


transport/tlscommon/tls.go
```diff
@@ -0,0 +1,246 @@
...
+
+		block, content = pem.Decode(content)
+		if block == nil {
+			if len(blocks) == 0 {
+				return nil, errors.New("no pem file")
+			}
+			break
+		}
+
+		if x509.IsEncryptedPEMBlock(block) { //nolint: staticcheck // deprecated, we have to get rid of it
```
cmacknz, line 101: Should we have an issue for these deprecated x509 functions?

CC @belimawr whose looking at improving the TLS code already.

cmacknz (approved):

transport/tlscommon/tls.go
belimawr: It seems [PEM encryption as specified in RFC 1423is insecure by design](https://pkg.go.dev/crypto/x509#IsEncryptedPEMBlock) rather than this specific function. So it's a bigger discussion as it means removing an functionality.

Anyway it's interesting discussion because in our case the key and the encrypted data live together...

PR #24, opened 2022-03-15
Move mage targets into a separate package and run for all platforms

rdner:
So, we can re-use these targets from other repositories.

Also, solved linting issues for Windows and Darwin.
Now linting is run for all 3 platforms on the CI in order to not miss issues.


magefile.go
```diff
@@ -21,238 +21,21 @@
...
-	"io"
-	"io/ioutil"
-	"log"
-	"net/http"
-	"os"
-	"path/filepath"
-	"strings"
-	"text/template"
-
+	// mage:import
```
rdner, line 24: This tag is required by the doc if mage targets are imported https://magefile.org/importing/

.ci/Jenkinsfile
```diff
@@ -38,6 +38,8 @@ pipeline {
           dir("${BASE_DIR}"){
             withMageEnv(){
               sh(label: 'Mage check', script: 'mage -v check')
+              sh(label: 'Lint Last Change on Darwin', script: 'GOOS=darwin mage -v llc')
+              sh(label: 'Lint Last Change on Windows', script: 'GOOS=windows mage -v llc')
```
rdner, line 42: I tried to run these as parallel steps but seems like the linter does not handle this well and randomly crashes. So, we have to make these steps sequential. However, they should not run for long because they lint only the change not the whole repository.

dev-tools/mage/deps.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
+)
+
+// Deps contains targets related to checking dependencies
+type Deps mg.Namespace
+
+// CheckModuleTidy checks if `go mod tidy` was run before the last commit.
+func (Deps) CheckModuleTidy() error {
```
kvch, line 14: What if instead of running `go mod` with `sh.Run` we used the package https://github.com/elastic/beats/tree/main/dev-tools/mage/gotool? I started porting a few calls because I need it for licence checking and for generating the NOTICE file.

rdner: I'm fine with that but I'd rather do it as a separate PR since this does not change the logic, it's just moving to a separate package.

kvch (approved):

PR #25, opened 2022-03-17
Move metrics

kvch:
This PR moves packages from `github.com/elastic/beats/v7/libbeat/metric` and its helpers to this lib.


cmacknz:
The changes here are kind of hard to review even with your suggestions on what to look at, because I don't get a real diff just the entire set of packages imported, and there are a lot of them. There are a few ways to make this easier to review. 

One would be to separate the changes into separate commits (one for the original import, one for changes on top of it).

Another would be to create a base branch that imports the packages with no changes, and then branch from that to make all the necessary edits. You'd need two PRs `changes -> base` and `base -> main` but it would show reviewers a nice change set.

metric/system/cgroup/reader.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil, fmt.Errorf("error finding cgroup version for pid %d: %w", pid, err)
+	}
+	if v == CgroupsV1 {
+		return r.GetV1StatsForProcess(pid)
+	}
+	return r.GetV2StatsForProcess(pid)
+}
+
+// GetV1StatsForProcess returns cgroup metrics and limits associated with a process.
+func (r *Reader) GetV1StatsForProcess(pid int) (*StatsV1, error) { // nolint: dupl // return value is different
```
cmacknz, line 182: This `dupl` warning is odd. This particular linter seems to have a high false positive rate.

fearful-symmetry:
Gonna give a +1 to @cmacknz 's suggestion of doing separate PRs for the import, then the needed changes. The huge green diff makes this a tad hard to review.

kvch:
I have force pushed the branch. The second commit contains the changes you should review: https://github.com/elastic/elastic-agent-libs/pull/25/commits/dff20022b8814f7d5b227622ff45ce753ddac688

Let me know if you need me to decrease the changeset even more.

kvch:
In the meantime I will try to figure out how to fix the linting errors.

metric/system/diskio/diskstat.go
```diff
@@ -48,6 +48,7 @@ type IOMetric struct {
 }
 
 // IOStat carries disk statistics for all devices
+// nolint: nolintlint // platform diffs
```
cmacknz, line 51: `nolint: nolintlint` this doesn't seem like a typo?

cmacknz:
Thanks for splitting this up, still a lot of changes but at least I can tell what they are.

I will let @fearful-symmetry do the approving here as they are more likely to spot problems in the details of what was changed.

cmacknz:
Once we have finished importing everything into elastic-agent-libs we should search for which linters we have disabled the most frequently and evaluate if we want to keep them.

metric/system/diskio/diskstat.go
kvch: No, it's not a typo. I have to turn off warnings for `nolint` for the next line because on Darwin the `structcheck` linting does not apply.

fearful-symmetry:
Huh, github does NOT like when you comment on that specific review. weird.

fearful-symmetry:
Bah, struggling to actually review/comment on the commits that have the changes. There's a few ignored error returns in `proc_linux_common` That I assume were added rote, but we should probably check.

Other than that, most of the changes seem fine? However, I'm still curious what the rationale is for moving away from `errors.Wrap` ? I assume there is one, and I just haven't read it.

cmacknz:
> I'm still curious what the rationale is for moving away from errors.Wrap ? I assume there is one, and I just haven't read it.

https://github.com/pkg/errors isn't maintained anymore, the repository is archived. It considers itself replaced by the built in https://pkg.go.dev/errors

PR #30, opened 2022-03-23
Linting test

rdner:
Just to test that the linting on Github Actions works.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -46,7 +46,7 @@ jobs:
           version: v1.44.2
 
           # Give the job more time to execute.
-          args: "--timeout=15m --whole-files"
+          args: --timeout=15m --whole-files
 
           # Optional: if set to true then the action will use pre-installed Go.
           skip-go-installation: true
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -46,7 +46,7 @@ jobs:
           version: v1.44.2
 
           # Give the job more time to execute.
-          args: --timeout=15m --whole-files
+          args: "--timeout=15m --whole-files"
 
           # Optional: if set to true then the action will use pre-installed Go.
           skip-go-installation: true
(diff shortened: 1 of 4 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -53,3 +53,9 @@ jobs:
```
andrewkroh, line 53: To match the way `mage llc` invokes it, try `--whole-files` to the args above. The GH action runs the linter using `--new-from-patch` and grabs the PR patch (e.g. https://github.com/elastic/elastic-agent-libs/pull/30.patch).

andrewkroh: Yep, I'm just trying to see what can be done to match what `mage llc` is doing in order to have annotations.

But maybe I should be thinking about it the other way. Like can we run `mage llc` in a Github Action and have it output the results in the annotation [format](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message) that Github recognizes. It appears this can be done natively by golangci-lint with `--out-format=github-actions`.

rdner: As you said, adding `--whole-files` basically matches the command and it does exactly the same just using the patch file instead of the revision. I'd vote for keeping the official Github action and using `--whole-files` with it. If we find something else we can always create our custom Github action but I don't it's worth it at this point.

rdner:
Just to summarise what happened in this PR:

* I pushed a new branch with 2 commits, both of them had issues to report but the Github Action run was completely successful (https://github.com/elastic/elastic-agent-libs/actions/runs/2028371049), however our `mage llc` target found 2 issues locally
* I pushed another commit with a bad dependency and the Github Action found only the last dependency issue https://github.com/elastic/elastic-agent-libs/runs/5660319116 but not the 2 previously introduced issues
* I tried to disable both `pkg` and `build` cashing in the Github Action but it didn't change anything and we still don't have 2 initial issues found by the Github action.
* I added `--whole-files` as a linter argument in the Github action and the linter reported all the issues https://github.com/elastic/elastic-agent-libs/runs/5664066262?check_suite_focus=true

So, the test showed that we have the same issue with the linting Github action as we solved with `mage llc` – we have to lint the whole files, otherwise all or some issues are not being reported and it's very unreliable.

PR #36, opened 2022-04-07
Extract function `SyncParent` to reuse in elastic agent 

narph:
Related to https://github.com/elastic/elastic-agent/issues/98

We need to sync parent after file is synced, this does not happen in function https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/storage/replace_store.go#L107

### DIFF
--- a/file/helper_aix.go
+++ b/file/helper_aix.go
@@ -24,7 +24,6 @@ import (
 
 // SafeFileRotate safely rotates an existing file under path and replaces it with the tempfile
 func SafeFileRotate(path, tempfile string) error {
-	parent := filepath.Dir(path)
 
 	if e := os.Rename(tempfile, path); e != nil {
 		return e
@@ -35,6 +34,14 @@ func SafeFileRotate(path, tempfile string) error {
 	// contain the new file being rotated in.
 	// On AIX, fsync will fail if the file is opened in read-only mode,
 	// which is the case with os.Open.
+	return SyncParent(path)
+
+}
+
+// SyncParent fsyncs parent directory
+func SyncParent(path string) error {
+	parent := filepath.Dir(path)
+
 	f, err := os.OpenFile(parent, os.O_RDWR, 0)
 	if err != nil {
 		return nil // ignore error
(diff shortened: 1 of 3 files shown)


blakerouse (approved):
Extraction looks good.

ph (changes requested):
Your implementation LGTM.

But there is something that bugs me in the files:

- If you look at the `file/helper_windows.go` the name of the file mention windows but there is no build tag defined.
- If you look at `file/helper_other.go` there twice the build tags, we should clean it up.
- If you look at `file/helper_aix.go` There is no build tag and I think we should remove it since we do not support officially aix.

I think we should clean the files and rename them appropriately to really understand what files is actually used in the build, for me it's unclear. Unless I've missed a feature in Go that automatically picks up the code based on the file suffix?

narph:
> 

same issues with `fileinfo-*` and `stderr-*` in the directory, @kvch what is the plan for the those files, can we make the same changes recommended by @ph  to all files?

kvch:
> If you look at file/helper_other.go there twice the build tags, we should clean it up.

FYI, the build tags are there twice because there was a change in the format of build tags. We did not remove the old format, so older Golang versions can be used. I am fine with removing it.

> I think we should clean the files and rename them appropriately to really understand what files is actually used in the build, for me it's unclear. Unless I've missed a feature in Go that automatically picks up the code based on the file suffix.

Golang indeed decides what files to build based on file suffixes: https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

I am fine with bringing consistency to the files. However, my concern is enforcing the consistency. It is great that it bugs some people, but not everyone has the same desire for consistency. We can spend a few minutes/hours on making sure build tags/file suffixes are the same, but it will be all in vain if a new person comes who does not have the same internal goal or does not spend enough time in the codebase for realizing that there is a pattern. Can we enforce it somehow? Maybe the linter already has a checker for this we could enable?

narph:
@ph , @kvch can we take this separately in a different issue, would be great to take this fix in as soon as we can

ph (approved):
Agree, lets create an issue to clean that up.

PR #38, opened 2022-04-12
Add notice generator target to `mage` and a NOTICE.txt file

kvch:
This PR adds a notice generator to the project. The generator can be used from other modules as well.

The PR appears big because of the size of the NOTICE file. I also moved the `gotool` helper package from beats. We will need it in the future for checking license headers, too.


dev-tools/templates/notice/overrides.json
```diff
@@ -0,0 +1,17 @@
...
+{"name": "github.com/xeipuuv/gojsonreference", "licenceFile": "LICENSE-APACHE-2.0.txt", "licenceType": "Apache-2.0"}
+{"name": "github.com/xeipuuv/gojsonschema", "licenceFile": "LICENSE-APACHE-2.0.txt", "licenceType": "Apache-2.0"}
+{"name": "github.com/chzyer/logex", "licenceType": "MIT"}
+{"name": "github.com/munnerz/goautoneg", "licenceType": "BSD-3-Clause"}
+{"name": "github.com/pelletier/go-buffruneio", "licenceType": "MIT"}
+{"name": "github.com/urso/magetools", "licenceType": "Apache-2.0"}
+{"name": "kernel.org/pub/linux/libs/security/libcap/cap", "licenceType": "BSD-3-Clause", "note": "dual licensed as GPL-v2 and BSD"}
+{"name": "kernel.org/pub/linux/libs/security/libcap/psx", "licenceType": "BSD-3-Clause", "note": "dual licensed as GPL-v2 and BSD"}
+{"name": "github.com/awslabs/kinesis-aggregation/go", "licenceType": "Apache-2.0", "url": "https://github.com/awslabs/kinesis-aggregation/blob/master/LICENSE.txt"}
+{"name": "github.com/pascaldekloe/goe", "licenceType": "CC0-1.0", "url": "https://github.com/pascaldekloe/goe/blob/master/LICENSE"}
```
rdner, line 17: Why do we have all of these? Is this file also generated or manually created?

kvch: It is created manually. `go-license-detector` is just guessing the type of the license based on license contents. However, some repos come with custom license text, so we need to tell the tool what is the type of the license.

rdner (approved):

PR #42, opened 2022-04-19
Move `github.com/elastic/beats/v7/libbeat/monitoring/report/buffer`

kvch:
In the first commit, the files are moved. Then the commit after that are the changes required by the checker and to be reusable.


michel-laterman (approved):
lgtm

PR #44, opened 2022-04-28
Move dialer helpers from `metricbeat/helper/dialer` to `transport` package

kvch:
It is required for removing the remaining Metricbeat dependencies from Agent.


rdner (approved):
I see no tests but I assume they have not been there before either.

PR #59, opened 2022-06-07
Remove empty string as default value for VerificationMode 

narph:
- Bug

## What does this PR do?

Original issue https://github.com/elastic/elastic-agent/issues/184
Elastic agent accepts empty string value for `ssl.verification_mode` (which is equivalent to `full`), however Endpoint is logging an error since it is not a supported option, but, it's also setting the verification mode to full, so verification modes match.

The PR is removing empty string as a default/supported verification mode option. Default mode will be `full`.

## Why is it important?

Avoids conflicts with other stakeholders.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/184


fearful-symmetry (approved):
Changes look good to me, although I'm assuming the change itself won't involve any user-facing breaking changes(?)

belimawr (changes requested):
@narph, the PR looks great, but to be on the safe side, I'd like to see test case for when the `verification_mode` is set to an empty string, e.g: test parsing a YAML like this:
```yaml
verification_mode: ""
```

narph:
> @narph, the PR looks great, but to be on the safe side, I'd like to see test case for when the `verification_mode` is set to an empty string, e.g: test parsing a YAML like this:
> 
> ```yaml
> verification_mode: ""
> ```

I've added some tests for this, including one with the scenario you suggested above but I assume by default `verification_mode` is not added in the configuration file instead of having a empty string value. Correct me if I am wrong here.

transport/tlscommon/types_test.go
```diff
@@ -20,6 +20,8 @@ package tlscommon
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
```
belimawr, line 24:
```suggestion
```

belimawr (approved):

PR #60, opened 2022-06-07
Synchronize lint rules across repositories

cmacknz:
I recently went through and disabled the most commonly ignored linters in beats: https://github.com/elastic/beats/pull/31683

This PR updates the lint configuration used in this repository with those that were removed in beats. Let's keep the lint rules consistent across our repositories.


kvch (approved):
Awesome as well!

PR #64, opened 2022-06-28
mapstr: optimyze Clone

florianl:
Reduce the ammount of allocations when using `Clone()`.
```
name     old time/op    new time/op    delta
Clone-8    8.95µs ± 0%    5.51µs ± 0%   ~     (p=1.000 n=1+1)

name     old alloc/op   new alloc/op   delta
Clone-8    4.99kB ± 0%    2.50kB ± 0%   ~     (p=1.000 n=1+1)

name     old allocs/op  new allocs/op  delta
Clone-8      64.0 ± 0%      33.0 ± 0%   ~     (p=1.000 n=1+1)
```

---------------------------------------------------------------------------------------------------------------------------

Type of change: `Enhancement`

## What does this PR do?

When profiling `filebeat` I have noticed heavy use of `MapStr.Clone()` operations that resulted in many `runtime.mallocgc()`. Looking at the code, I did see that the resulting `M` was not sized properly even if its resulting size is known in advance.

As the added benchmark demonstrates, there is a noticeable reduction of memory allocation even for small `MapStr` when using `Clone()` now.

## Why is it important?

Incremental resizing of `MapStr`, as it is done when `MapStr` is not properly sized in advance, will give the garbage collector more work to do. So this change does not only reduce the number of allocations but also reduces the work that needs to be done by the garbage collector.

## Checklist

      As there is only a minimal code change, I have skipped this one. Please let me know, if you want more comments.
- [x] I have added ~~tests~~ _benchmarks_ that prove my fix is effective or that my feature works
      For users of this package there is no "visible" change. So I did not add something.

### DIFF
--- a/mapstr/mapstr.go
+++ b/mapstr/mapstr.go
@@ -152,13 +152,13 @@ func (m M) CopyFieldsTo(to M, key string) error {
 // Clone returns a copy of the M. It recursively makes copies of inner
 // maps.
 func (m M) Clone() M {
-	result := M{}
+	result := make(M, len(m))
 
-	for k, v := range m {
-		if innerMap, ok := tryToMapStr(v); ok {
-			v = innerMap.Clone()
+	for k := range m {
+		if innerMap, ok := (m[k]).(M); ok {
+			result[k] = innerMap.Clone()
 		}
-		result[k] = v
+		result[k] = m[k]
 	}
 
 	return result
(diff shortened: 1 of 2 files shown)


cmacknz (approved):
Nice! There will need to be a follow up PR in beats to pick up this change, I can tag it after we merge this.

cmacknz:
Beats PR incorporating this change: https://github.com/elastic/beats/pull/32161

PR #76, opened 2022-08-23
kibana: don't try to wrap nil errors

efd6:
## What does this PR do?

This makes sure that only non-nil errors are wrapped into errors returned by function in the kibana client.

## Why is it important?

Non-nil error should not be included in error wrapping.

## Checklist

- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~I have added an entry in `CHANGELOG.md`~~

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/kibana/client.go
+++ b/kibana/client.go
@@ -305,12 +305,12 @@ func (client *Client) readVersion() error {
 
 	code, result, err := client.Connection.Request("GET", statusAPI, nil, nil, nil)
 	if err != nil {
-		return fmt.Errorf("HTTP GET request to %s/api/status fails: %w. Response: %s",
-			client.Connection.URL, err, truncateString(result))
+		return fmt.Errorf("HTTP GET request to %s/api/status fails: %w (status=%d). Response: %s",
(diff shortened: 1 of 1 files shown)


kibana/client.go
```diff
@@ -304,10 +304,14 @@ func (client *Client) readVersion() error {
 	}
 
 	code, result, err := client.Connection.Request("GET", statusAPI, nil, nil, nil)
-	if err != nil || code >= 400 {
+	if err != nil {
 		return fmt.Errorf("HTTP GET request to %s/api/status fails: %w. Response: %s",
 			client.Connection.URL, err, truncateString(result))
 	}
+	if code >= 400 {
+		return fmt.Errorf("HTTP GET request to %s/api/status fails. Response: %s",
```
cmacknz, line 312: Let's include the HTTP status code in the error here

cmacknz (approved):

efd6:
I don't know why the lint checks are failing here. It seems that a previous PR did not update the licenses.

PR #80, opened 2022-09-06
ci: change labels to match the OS version

v1v:
`orka` is deprecated in favour of `macos<version>`

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -58,7 +58,7 @@ pipeline {
         axes {
           axis {
             name 'PLATFORM'
-            values 'ubuntu-20.04 && immutable', 'windows-2019 && windows-immutable', 'darwin && orka && x86_64'
+            values 'ubuntu-20.04 && immutable', 'windows-2019 && windows-immutable', 'x86_64 && macos11'
           }
         }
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -69,7 +69,12 @@ pipeline {
                 unstash 'source'
                 dir("${BASE_DIR}"){
                   withGoEnv(){
-                    goTestJUnit(options: '-v ./...', output: 'junit-report.xml')
+                    whenTrue(isDarwin()) {
```
cmacknz, line 72: Can you add a comment explaining why there is a Darwin special case here? Otherwise LGTM.

cmacknz (approved):

.ci/Jenkinsfile
v1v:
```suggestion
                    // goTestSum for go1.18.5 on MacOS produces a segmentation violation. See https://github.com/elastic/elastic-agent-libs/pull/80
                    // For the time being there will not junit transformation in MacOS
                    whenTrue(isDarwin()) {
```

cmacknz: Thanks!

PR #82, opened 2022-09-13
Add SetLevel functionality to logp package

amirbenun:
## What does this PR do?
This PR exposes a global `logp.SetLevel` that changes the minimal level of logging for all of the loggers that were created through `logp.NewLogger`.
Ideally, it should be able to change the log level of a running agent without having to restart it.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/logp/core_test.go
+++ b/logp/core_test.go
@@ -57,7 +57,7 @@ func TestLogger(t *testing.T) {
 
 func TestLoggerLevel(t *testing.T) {
 	if err := DevelopmentSetup(ToObserverOutput()); err != nil {
-		t.Fatal(err)
+		t.Fatalf("cannot initialise logger on development mode: %+v", err)
 	}
 
 	const loggerName = "tester"
--- a/logp/core_test.go
+++ b/logp/core_test.go
@@ -96,6 +96,36 @@ func TestLoggerLevel(t *testing.T) {
 	}
 }
 
+func TestLoggerSetLevel(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmacknz:
I think this looks reasonable, thanks!

Have you tested that this works as expected in one of the Beats to ensure this works when integrated into a real application?

logp/core_test.go
```diff
@@ -96,6 +96,36 @@ func TestLoggerLevel(t *testing.T) {
 	}
 }
 
+func TestLoggerSetLevel(t *testing.T) {
+	if err := DevelopmentSetup(ToObserverOutput()); err != nil {
+		t.Fatal(err)
```
belimawr, line 101: [Suggestion]
It helps to add a message explaining what failed
```suggestion
		t.Fatalf("cannot initialise logger on development mode: %s", err)
```

belimawr (approved):
LGTM, but as @cmacknz mentioned, it's best to test it with a Beat before merging, just to be on the safe side.

amitkanfer:
@amirbenun can we push this forward?

amirbenun:
Update: Testing before merging was the right decision, as I see that calling `SetLevel` from cloudbeat crashes the app.
Looking into it.

amirbenun:
I was able to test the `SetLevel` functionality with cloudbeat.
When calling the following func:
Screen Shot 2022-09-22 at 18 08 45

I got the following logs:
Screen Shot 2022-09-22 at 18 09 04

PR #83, opened 2022-10-05
Make the NewX() monitoring variables safe to re-register

fearful-symmetry:
## What does this PR do?

Fixes https://github.com/elastic/beats/issues/33125

Because this problem makes me a tad paranoid, I decided to implement the fix on a more global level. Specifically, this changes the behavior of the `NewX()` monitoring functions so they don't panic when a variable of the same name and type is re-registered. This fix is a tad opinionated, and I'd like to hear input from others on a few things:

- Instead of panicking when a variable is re-registered, we return the existing variable. My concern is that this is a tad implicit, and code might be assuming that a call to `NewX()` will create a new monitoring variable with a zero value. Should we zero out the underlying metric, and then return?
- If someone tries to re-register a variable with options, we just revert to a panic. I figured this was the most understandable behavior, as certain options, such as expvar exporting, can't really be modified. Should we attempt to merge or reset variables where possible?

## Checklist

### DIFF
--- a/monitoring/metrics.go
+++ b/monitoring/metrics.go
@@ -20,6 +20,7 @@ package monitoring
 import (
 	"encoding/json"
 	"expvar"
+	"fmt"
 	"math"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


faec (approved):
This all seems reasonable. I think we probably don't want to reset a metric that has already been defined. However, can we still log a warning when things are redefined? It makes sense not to panic, but I think we should still try to remove redefinitions when we discover them, so I'd still rather have them show up in the logs.

fearful-symmetry:
@faec regarding logging, the main use case here is monitoring variables that get hit when elastic-agent reloads a beat component. So my thinking is that we would fill the logs with scary-looking but irrelevant log lines, but it might still be worth doing at a debug level?

PR #86, opened 2022-10-19
Support ssl.peer_name configuration

folbricht-stripe:
## What does this PR do?

Adds support for setting the peer certificate name in the config with `ssl.peer_name`. When set, this will configure the `TestTLSDialer` to use the given name rather than extracting it from the peer address.

The change is fully backwards-compatible.

## Why is it important?

Allows the use of endpoints for which the SAN name doesn't match the address used to connect to them.

## Checklist

- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [x] I have added an entry in `CHANGELOG.md`

## Related issues

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - dev-tools: added GoImports target #74
+- transport/tlscommon: New ssl.peer_name config option #86
 
 ### Changed
 
--- a/transport/tls.go
+++ b/transport/tls.go
@@ -51,9 +51,13 @@ func TestTLSDialer(
 			return nil, fmt.Errorf("unsupported network type %v", network)
 		}
 
-		host, _, err := net.SplitHostPort(address)
-		if err != nil {
-			return nil, err
+		host := config.PeerName
+		if host == "" {
+			var err error
+			host, _, err = net.SplitHostPort(address)
+			if err != nil {
+				return nil, err
+			}
 		}
 
 		var tlsConfig *tls.Config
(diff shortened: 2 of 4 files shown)


rdner:
Hi @folbricht-stripe and thank you for your contribution.

Could you explain a particular use-case for this change? SAN is supposed to be extracted from the certificate and validated against the target host. Why is it troublesome to include SAN in the certificate?

We have a guide for generating certificates that are supposed to be used with our services here https://www.elastic.co/guide/en/fleet/current/secure-connections.html

belimawr:
> In our case, for a number of reasons, there's a kind of proxy/proxy in between the client and the server (one that does not terminate TLS). So the certificate presented by the client does not match the public DNS name for the filebeat server endpoint.
> 
> I do realize this is an unusual setup and this config option will rarely be needed, but it's also quite safe as it allows setting fixed SAN names in config where necessary, adding a little more flexibility.

If I understood correctly, what you need is certificate validation without hostname validation, it that correct? For that you can set `verification_mode: certificate`: https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#client-verification-mode.

When using proxies, ideally the proxy would handle TLS and the certificates would be valid for the proxy hostname.

That is a very sensitive piece of code, I rather not add this exception because the problem can be solved in other ways.

folbricht-stripe:
Ideally, it'd be good to have certificate + name validation but with a fixed (configured) name. Using `verification_mode: certificate` would work, but it's slightly less strict since it's missing the name comparison.

folbricht-stripe:
Worked around it and only using `verification_mode: certificate` now, so this can be closed.

PR #88, opened 2022-11-05
Add Core function to logp.Logger to allow access to backing zapcore.Core.

blakerouse:
## What does this PR do?

Adds the ability to retrieve the `zapcore.Core` from the `logp.Logger`. This allows the ability to write entries directly into the logger bypassing any log levels settings.

## Why is it important?

This is needed for the work on https://github.com/elastic/elastic-agent/issues/221. The `io.Writer` connected to `stdout` and `stderr` from the spawned subprocesses will read the output from the process and output it directly to the `zapcore.Core`. That is required because each component and unit can have its own log level different from the Elastic Agent log level. Because of that design it requires that log messages bypass the log level settings for the logger used by Elastic Agent.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/221

### DIFF
--- a/logp/logger.go
+++ b/logp/logger.go
@@ -224,6 +224,11 @@ func (l *Logger) Sync() error {
 	return l.logger.Sync()
 }
 
+// Core returns the backend zapcore.Core for the logger.
+func (l *Logger) Core() zapcore.Core {
+	return l.logger.Core()
+}
+
 // L returns an unnamed global logger.
 func L() *Logger {
 	return loadLogger().logger


logp/logger.go
```diff
@@ -224,6 +224,11 @@ func (l *Logger) Sync() error {
 	return l.logger.Sync()
 }
 
+// Core returns the backend zapcore.Core for the logger.
+func (l *Logger) Core() zapcore.Core {
```
rdner, line 228: I wonder if it would be better to define a type alias like it's done with the `zap.Option`:

https://github.com/elastic/elastic-agent-libs/blob/89215adc2946dcd3cfc873fb954c51eb7fb8fd91/logp/logger.go#L27-L28

I don't know how far we want to go exposing the `zap` types in our library. I see that the `Logger` struct just has `zap` types directly:

https://github.com/elastic/elastic-agent-libs/blob/89215adc2946dcd3cfc873fb954c51eb7fb8fd91/logp/logger.go#L30-L34

rdner (approved):

logp/logger.go
rdner: It would rather an abstraction should we decide to switch from zap to something else. I guess that's the point. But I'm okay with this.

PR #91, opened 2022-12-05
Migrate loader library from elastic-agent to here

fearful-symmetry:
## What does this PR do?

This PR doesn't really do anything on its own, but is a dependency on an upcoming PR that will give `elastic-agent-client` a standalone mode, allowing a client to run from a raw elastic-agent.yml config file, without an attached instance of elastic-agent. This library was previously in elastic-agent's internal library, but it's very useful for any code that needs to read a yaml file that's similarly structured.

## Why is it important?

This will be a requirement for a PR to improve elastic-agent-client.

## Checklist


fearful-symmetry:
Code Analysis failure is related to the code duplication on the tests, and issues outside the scope of this PR.

faec (approved):

PR #93, opened 2023-01-09
Fix possible private key debug logs

cmacknz:
Often these are a file path, but occasionally they can be a private key
embedded directly into the YAML configuration. Remove logging of the key
to avoid dumping private keys into our logs when debug logging is enabled.

Our own documentation has an example of a private key being embedded pasted directly into the YAML for this field which would be pasted verbatim into our debug logs without being sanitized first: https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#client-key


transport/tlscommon/tls.go
```diff
@@ -66,7 +66,7 @@ func LoadCertificate(config *CertificateConfig) (*tls.Certificate, error) {
 		return nil, err
 	}
 
-	log.Debugf("Loading certificate: %v and key %v", certificate, key)
+	log.Debugf("Loading certificate: %v", certificate)
```
cmacknz, line 69: I could remove the certificate logging too if we think it isn't actually useful. We can usually get this by asking for the configuration used to run Beats/Agent/whatever instead although this is more convenient and won't be wrong.

leehinman (approved):
LGTM

transport/tlscommon/tls.go
belimawr: I agree with removing it, either here or in another PR. What I believe would be more helpful is to log (as debug) if we're loading a file (and its path) or reading the cert directly from the config.

belimawr (approved):

transport/tlscommon/tls.go
leehinman: I have a small preference for continuing to log the cert.  Because it is a public key, there shouldn't be any security concerns in logging it and it can make it easy to see things like mistakes in encoding etc.  That being said I'd be OK if it went away.

I really like the idea of logging the location of the key material.

cmacknz: There is a helper for detecting a PEM encoded string, so using that I can keep logging the location of the key when it is a file.

cmacknz: Also I agree logging the cert isn't an issue.

cmacknz:
I have also decided to remove the CHANGELOG.md file maintained in this repository as a bonus change. I went to update it, realized it was out of date, and also realized that for this repository it wasn't doing anything that the commit log isn't already doing for us.

transport/tlscommon/tls.go
```diff
@@ -66,7 +66,14 @@ func LoadCertificate(config *CertificateConfig) (*tls.Certificate, error) {
 		return nil, err
 	}
 
-	log.Debugf("Loading certificate: %v and key %v", certificate, key)
+	// Do not log the key if it was provided as a string in the configuration to avoid
+	// leaking private keys in the debug logs. Log when the key when it is a file path.
+	if IsPEMString(key) {
+		log.Debugf("Loading certificate: %v with key from PEM string in config", certificate)
+	} else {
+		log.Debugf("Loading certificate: %v and key %v", certificate, key)
```
leehinman, line 74: I think the key snuck back in the debug log.

leehinman (changes requested):

transport/tlscommon/tls.go
cmacknz: This should only log the key if `IsPEMString(key)` is false, which essentially detects whether the key contents are a string in the config and allows logging the path we are using when loading it from a file.

I can remove this entirely if we are concerned about it.

leehinman: 🤦 ok, so it should be a file path at that point which is fine.

leehinman (approved):

transport/tlscommon/tls.go
```diff
@@ -66,7 +66,14 @@ func LoadCertificate(config *CertificateConfig) (*tls.Certificate, error) {
 		return nil, err
 	}
 
-	log.Debugf("Loading certificate: %v", certificate)
+	// Do not log the key if it was provided as a string in the configuration to avoid
+	// leaking private keys in the debug logs. Log when the key when it is a file path.
```
belimawr, line 70: A small typo:
```suggestion
	// leaking private keys in the debug logs. Log when the key is a file path.
```

belimawr (approved):

PR #101, opened 2023-02-14
fix: update macos_service environment flag value

kruskall:
## What does this PR do?

Update macos_service value for the flag `--environment`.
Passing `--environment macos_service` will always fail because the equality check is comparing against `macOS_service`. Using the new name does not help because the value is lowercased before the comparison.

## Why is it important?

This is breaking an APM Server flag.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Related to https://github.com/elastic/apm-server/pull/10250

### DIFF
--- a/logp/environment.go
+++ b/logp/environment.go
@@ -47,7 +47,7 @@ const (
 	defaultEnvironmentString        = "default"
 	systemdEnvironmentString        = "systemd"
 	containerEnvironmentString      = "container"
-	macOSServiceEnvironmentString   = "macOS_service"
+	macOSServiceEnvironmentString   = "macos_service"
 	windowsServiceEnvironmentString = "windows_service"
 	invalidEnvironmentString        = "<invalid>"
 )


logp/environment.go
```diff
@@ -47,7 +47,7 @@ const (
 	defaultEnvironmentString        = "default"
 	systemdEnvironmentString        = "systemd"
 	containerEnvironmentString      = "container"
-	macOSServiceEnvironmentString   = "macOS_service"
+	macOSServiceEnvironmentString   = "macos_service"
```
cmacknz, line 50: This was copied from Beats, the original code is still used in 7.17:

https://github.com/elastic/beats/blob/9ded7059eb086d009d584266e83cf82306502516/libbeat/logp/environment.go#L46-L83

```go
// String returns the string representation the configured environment
func (v Environment) String() string {
	switch v {
	case DefaultEnvironment:
		return "default"
	case SystemdEnvironment:
		return "systemd"
	case ContainerEnvironment:
		return "container"
	case MacOSServiceEnvironment:
		return "macOS_service"
	case WindowsServiceEnvironment:
		return "windows_service"
	default:
		return "<invalid>"
	}
}

// ParseEnvironment returns the environment type by name.
// The parse is case insensitive.
// InvalidEnvironment is returned if the environment type is unknown.
func ParseEnvironment(in string) Environment {
	switch strings.ToLower(in) {
	case "default":
		return DefaultEnvironment
	case "systemd":
		return SystemdEnvironment
	case "container":
		return ContainerEnvironment
	case "macos_service":
		return MacOSServiceEnvironment
	case "windows_service":
		return WindowsServiceEnvironment
	default:
		return InvalidEnvironment
	}
}
```

It looks like we used both versions before and just arbitrarily picked one here. The important one seems to be the parse case which uses the lowercase spelling.

AFAIK nobody has complained about this as a bug so it is unlikely to be widely used, otherwise I'd suggest accepting both versions. Leaving it like this should be fine.

cmacknz (approved):

logp/environment.go
kruskall: This PR would still accept both versions (`macOS_service` and `macos_service`). 
The issue is that the current code (and APM Server) returns an error for both.

The current code expects `strings.ToLower(in) == "macOS_service"` which will never be true.

https://github.com/elastic/elastic-agent-libs/blob/fa8f242aaa3ef1292b29ee7761d7b9a083b4e04c/logp/environment.go#L76-L91

cmacknz: Ah missed the ToLower conversion, thanks for clarifying. This indeed has never worked.

cmacknz:
I added the apm-server team to this repository with maintainer permissions, you should be able to merge and create tags. Let me know if something prevents you from doing that.

kruskall:
@cmacknz Thank you for the help! :bow:

kruskall:
Are you okay with releasing v0.3.1 ?

PR #108, opened 2023-03-21
Adding convenience method for FQDN-aware hostname

ycombinator:
## What does this PR do?

This PR aliases the `github.com/elastic/go-sysinfo/types/HostInfo` type to `github.com/elastic/elastic-agent-libs/sysinfo/HostInfo` and adds a convenience method for reporting the FQDN-aware hostname.

## Why is it important?

To consolidate logic of reporting the appropriate hostname in a single place.

Also see discussion on https://github.com/elastic/go-sysinfo/pull/156#discussion_r1143910878.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-


sysinfo/types/host.go
```diff
@@ -0,0 +1,32 @@
...
+
+package types
+
+import "github.com/elastic/go-sysinfo/types"
+
+type HostInfo types.HostInfo
+
+// FQDNAwareHostname returns the system hostname, honoring the given
+// flag to report the FQDN or not.
+func (host HostInfo) FQDNAwareHostname(wantFQDN bool) string {
```
rdner, line 26: I don't think we need this function:
1. We can access the fields directly
2. This function has 2 behaviour based on a boolean flag, I'd avoid this patter, it's better to have 2 functions then. But as I mentioned above, those 2 function would be just unnecessary getter functions for the accessible/public fields.

ycombinator: Funny, because I received exactly the opposite feedback in a very similar situation on another (related to FQDN) PR 😄:

https://github.com/elastic/beats/pull/34456#discussion_r1130675452

Without a helper function like this, we will repeat the (trivial) conditional logic in a few places: https://github.com/elastic/elastic-agent/pull/2218/commits/d2c051221b939b3d1234920cb98271cb17ccbf6a

> This function has 2 behaviours based on a boolean flag, I'd avoid this pattern, it's better to have 2 functions then. 

Can you elaborate a bit on _why_ this pattern should be avoided?   What is the downside or "cost" of such a pattern?

I think helper functions like this one are not uncommon and it's fine to have them as long as they're documented and tested well.  It helps avoid repeating the same logic in multiple places and risking getting it wrong.  I know the logic is fairly trivial in this case but I'm not really sure what the harm is in having a helper function encapsulate it.  And, of course, the underlying public fields are still available for direct access in situations where that would make more sense.

sysinfo/types/host.go
```diff
@@ -0,0 +1,39 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package types
+
+import "github.com/elastic/go-sysinfo/types"
+
+type HostInfo types.HostInfo
```
andrewkroh, line 22: I don't think it needs to alias the type to achieve the result of having a helper. You could have a standalone function that accepts a `types.HostInfo` as a parameter (e.g. `hostname.Get(types.HostInfo, mode hostname.Mode) string`).

sysinfo/types/host.go
rdner: > Can you elaborate a bit on why this pattern should be avoided? What is the downside or "cost" of such a pattern?

sure, for example, when you see 

```go
FQDNAwareHostname(true)
```

you cannot tell what `true` means without going to the function definition. That's why boolean flags as function arguments are commonly avoided. It's better to predefine constants (like a enum) and use them instead. A constant name is way more descriptive than a boolean.

So, it could become something like:

```go
type HostnameMode int

const (
	FQDN HostnameMode = 0
	Short HostnameMode = 1 // this might have a better name
)

func Hostname(mode HostnameMode) string
```

and then it can be used like:

```go
hn := Hostname(FQDN)
```

If we don't want to have a type alias, it's:

```go
func Hostname(hostInfo types.HostInfo, mode HostnameMode) string
```

ycombinator:
With recent changes in how the `go-sysinfo` library returns FQDN, this convenience method no longer makes sense.  Closing this PR without merging.

PR #110, opened 2023-03-22
Add target for go benchmarks and benchstat

alexsapran:
## What does this PR do?

This PR adds 2 new mage targets for running the go native benchmark tests and running the benchstat for parsing the results.

With the following targets, you can do the following

- Run benchmarks and print or save the output
```bash
$ mage benchmark:run
```

This will execute the go benchmark.
You can control the number of iterations that the benchmark will execute by defining an env var `BENCH_COUNT`
You can write the output into a file by defining the `OUTPUT` variables.

Example:
```bash
$ OUTPUT=base.txt BENCH_COUNT=1 mage benchmark:run
```

When it is finished there is going to be a file `base.txt` with the output of the executed benchmarks.

- Examining the benchmark output

```bash
$ BASE=base.txt mage benchmark:diff
```

This will parse and print the output of the benchmark.

To take the most out of the diff is when comparing benchmark results.

Assuming that you have executed the `benchmark:run` before you start performing your changes and have the output stored in the `base.txt` then after you have completed your changes you can re-run the same benchmark command, but this time writes it to `next.txt`. Then you can compare the outcome by the following command.

```bash
$ BASE=base.txt NEXT=next.txt mage benchmark:diff
```

Defining an OUTPUT variable in the above will save the comparison in the defined output file.

## Why is it important?

Those new targets are going to be used as part of future work and be integrated into the CI workflow of this repository.

### DIFF
--- a/dev-tools/mage/benchmark.go
+++ b/dev-tools/mage/benchmark.go
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 package mage
 
 import (
--- a/dev-tools/mage/benchmark.go
+++ b/dev-tools/mage/benchmark.go
@@ -18,7 +18,6 @@
 package mage
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -31,6 +30,7 @@ import (
 
 	"github.com/elastic/elastic-agent-libs/dev-tools/mage/gotool"
 	"github.com/magefile/mage/mg"
+	"github.com/magefile/mage/sh"
 )
 
 const (
@@ -53,8 +53,9 @@ func (Benchmark) Deps() error {
 	return nil
 }
 
-// Run execute the go benchmark tests for this repository, define OUTPUT to write results into a file
-func (Benchmark) Run(ctx context.Context) error {
+// Run execute the go benchmark tests for this repository, by defining the variable OUTPUT you write the results
+// into a file. Optional you can set BENCH_COUNT to how many benchmark iteration you want to execute, default is 8
+func (Benchmark) Run() error {
 	mg.Deps(Benchmark.Deps)
 	log.Println(">> go Test: Benchmark")
 	outputFile := os.Getenv("OUTPUT")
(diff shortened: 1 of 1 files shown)


dev-tools/mage/benchmark.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return nil
+}
+
+// Run execute the go benchmark tests for this repository, define OUTPUT to write results into a file
+func (Benchmark) Run(ctx context.Context) error {
+	mg.Deps(Benchmark.Deps)
+	log.Println(">> go Test: Benchmark")
+	outputFile := os.Getenv("OUTPUT")
+	benchmarkCountOverride := os.Getenv("BENCH_COUNT")
```
fearful-symmetry, line 61: Mention `BENCH_COUNT` in the godoc perhaps?

dev-tools/mage/benchmark.go
```diff
@@ -0,0 +1,170 @@
...
+	case goTestErr == nil:
+		return nil
+	case errors.As(err, &goTestErr):
+		return fmt.Errorf("failed to execute go test -bench command: %w", err)
+	default:
+		return fmt.Errorf("failed to execute go test -bench command %w", err)
+	}
+}
+
+// Diff compare 2 benchmark outputs, Required BASE variable for parsing results, define NEXT to compare base with next and optional OUTPUT to write to file
```
fearful-symmetry, line 91: The godoc here is a little confusing at first, maybe mention that `BASE` and `NEXT` are environment variables?

dev-tools/mage/benchmark.go
```diff
@@ -0,0 +1,170 @@
...
+	default:
+		return fmt.Errorf("failed to execute benchstat command!! %w", err)
+	}
+
+}
+
+// runCommand is executing a command that is represented by cmd.
+// when defining an outputFile it will write the stdErr, stdOut of that command to the output file
+// otherwise it will capture it to stdErr, stdOut of the console used.
+func runCommand(ctx context.Context, env map[string]string, cmd string, outputFile string, args ...string) (*exec.Cmd, error) {
```
fearful-symmetry, line 128: Is there a reason we're not using the mage [sh](https://pkg.go.dev/github.com/magefile/mage@v1.14.0/sh) libraries here? Might remove some of the yak shaving.

alexsapran: Great suggestion, I really don't know why. I recall I tried to use it but I had some issues that I cannot reproduce anymore, so the problem was unrelated to the use of `sh` that pushed me to use the `exec` directly.

dev-tools/mage/benchmark.go
alexsapran: I tried to write it a bit differently, please let me know if it's better now.

dev-tools/mage/benchmark.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return nil
+}
+
+// Run execute the go benchmark tests for this repository, by defining the variable OUTPUT you write the results
+// into a file. Optional you can set BENCH_COUNT to how many benchmark iteration you want to execute, default is 8
+func (Benchmark) Run() error {
+	mg.Deps(Benchmark.Deps)
+	log.Println(">> go Test: Benchmark")
+	outputFile := os.Getenv("OUTPUT")
```
leehinman, line 61: just a suggestion.  could we have a default for output that was something like "benchmark-5e71fe7a79.txt", where the 5e71fe7a79 is the output from something like `git log -1 --pretty=format:%h`.  That would make it easy to keep benchmarks from releases and when you are working on code would make it easy to see which benchmark goes with what state of the code.

leehinman (approved):
LGTM

fearful-symmetry (approved):

dev-tools/mage/benchmark.go
alexsapran: Initially, I did not want to write a file every time but revisiting now seems like a good suggestion. 
I would like to get this merged and I will do follow-up work and will incorporate your suggestion then.

PR #112, opened 2023-03-23
Fix ReportBool in visitor, add comments

fearful-symmetry:
## What does this PR do?

This is a little one-liner fix, the `ReportBool` method would report a string, which would result in a bool metric value becoming `"bool_value":"bool_value"` once reported as JSON.

## Why is it important?

This is a bug.

## Checklist

### DIFF
--- a/monitoring/metrics_test.go
+++ b/monitoring/metrics_test.go
@@ -38,6 +38,30 @@ func TestSafeVars(t *testing.T) {
 	require.Equal(t, uint64(5), testSecondUint.Get())
 }
 
+func TestVarsTypes(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmacknz:
Is this covered by the tests in https://github.com/elastic/elastic-agent-libs/blob/main/monitoring/visitor_expvar_test.go? Can it be? Seems like something a simple test would catch.

fearful-symmetry:
Yah, good point, let me see if I can add a test somewhere...

leehinman (approved):

cmacknz (approved):
Thanks for the test!

PR #121, opened 2023-05-08
feat: bump default TLS version to TLS v1.2

kruskall:
## What does this PR do?

Bump the default TLS version to TLS v1.2 from TLS v1.1.

Support for TLS v1.1 is not removed for the sake of compatibility.

## Why is it important?

Motivated from the work in APM Server toward bumping the default TLS version to 1.2: https://github.com/elastic/apm-server/issues/10491

TLS v1.1. was formally deprecated in 2021 and is considered obsolete: https://datatracker.ietf.org/doc/html/rfc8996

Several standard and the NIST guidelines for TLS implementation recommends not using TLS 1.1 or lower.

## Checklist

~- [ ] I have commented my code, particularly in hard-to-understand areas~
~- [ ] I have added an entry in `CHANGELOG.md`~

## Author's Checklist

- [ ]

## Related issues

Related to https://github.com/elastic/apm-server/issues/10491

### DIFF
--- a/transport/tlscommon/versions_default.go
+++ b/transport/tlscommon/versions_default.go
@@ -31,7 +31,7 @@ const (
 
 var (
 	// TLSVersionMin is the min TLS version supported.
-	TLSVersionMin = TLSVersion12
+	TLSVersionMin = TLSVersion11
 
 	// TLSVersionMax is the max TLS version supported.
 	TLSVersionMax = TLSVersion13


belimawr (approved):

cmacknz:
Let's hold off on merging this until we have have sanity checked that this won't do anything surprising when Beats, Elastic Agent, and Fleet Server are eventually updated with this change. 

In particular we need to verify that an Elastic Agent defaulting to TLS v1.2 can connect to a Fleet Server defaulting to TLS v1.1 and vice versa. The version negotiation should just work, but the downside and impact of being wrong about that is big enough that I don't want to assume there will be no problems. Let's actually test it first.

cmacknz:
No progress on this, not a high priority for us yet.

jlind23:
@pierrehilbert can we please make this a priority for one of the next upcoming sprints? As stated by @cmacknz we should test this and it will most probably work OOTB.
Testing 7.17.X and 8.X together would be the path to take.

ycombinator:
As mentioned in https://github.com/elastic/elastic-agent-libs/pull/121#issuecomment-1538889426, the work left to do on this PR now is [testing the impact]() of the changes in this PR before we can be confident about merging it.  Concretely, the following two tests will need to be performed:

### Agent using TLSv1.1 can talk to Fleet Server using TLSv1.2

For this test, Elastic Agent will need to built without any changes to its dependencies.  This will produce an Agent that's using TLSv1.1.

Fleet Server will need to be built with it's `elastic-agent-libs` dependency temporarily pointing to a local copy that has this PR checked out in it.  This will produce a Fleet Server that's using TLSv1.2.

Then it's a matter of enrolling the built Agent with the built Fleet Server and smoke testing that a few common operations in Fleet that would cause back-and-forth communication between Agent and Fleet Server work as expected.

### Agent using TLSv1.2 can talk to Fleet Server using TLSv1.1

For this test, Elastic Agent will need to built with it's `elastic-agent-libs` dependency temporarily pointing to a local copy that has this PR checked out in it.  This will produce an Agent that's using TLSv1.2. 

Fleet Server will need to be built without any changes to its dependencies.  This will produce a Fleet Server that's using TLSv1.1.

Then it's a matter of enrolling the built Agent with the built Fleet Server and smoke testing that a few common operations in Fleet that would cause back-and-forth communication between Agent and Fleet Server work as expected.

michel-laterman (approved):
We've added `tlscommon.SetInsecureDefaults()` so that users can restore previous behaviour (on 8.x branches).

This is already being used in beats, elastic-agent, and fleet-server

PR #128, opened 2023-06-02
monitoring - Rename NewRegistry to MustNewRegistry

andrewkroh:
## What does this PR do?

`registry.NewRegistry` could panic and it was not documented. This is a breaking change to force callers of this method to evaluate if they are willing to accept the current behavior which panics.

This creates two choices for existing users. The can use either:

    MustNewRegistry(...) *Registry      // Keep the same behavior.
    NewRegistry(...) (*Registry, error) // Handle the error.

A similar change was made to the `registry.Add` method (except existing code will continue to compile since it previously had no return value).

## Why is it important?

In Beats there have been several bugs that resulted in a panic. I would rather have an option to receive an error and handled it as I see fit rather than have a panic forced on my application.

## Checklist

- [ ] ~I have added an entry in `CHANGELOG.md`~ _What changelog file? This does not exist._

### DIFF
--- a/monitoring/adapter/go-metrics.go
+++ b/monitoring/adapter/go-metrics.go
@@ -36,9 +36,9 @@ import (
 // the go-metrics.Registry interface.
 //
 // Note: with the go-metrics using `interface{}`, there is no guarantee
-//       a variable satisfying any of go-metrics interfaces is returned.
-//       It's recommended to not mix go-metrics with other metrics types
-//       in the same namespace.
+// a variable satisfying any of go-metrics interfaces is returned.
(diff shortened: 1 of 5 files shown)


andrewkroh:
https://github.com/search?q=org%3Aelastic+path%3A*.go+%2F%5C.NewRegistry%5C%28%2F&type=code&p=1

After taking a look at how many usages of `NewRegistry` there are, I'm going to close this and add a non-breaking implementation. I still think this would be the best route from a clear API perspective, but I don't want to have to update all the current usages.

PR #130, opened 2023-06-08
Fix logging of matching CA certificate

jsoriano:
## What does this PR do?

When the 'ca_trusted_fingerprint' matches one of the certificates provided, log success only if the matching certificate is actually a CA, otherwise log that the matching certificate is not a CA.

With current code you can find contradictory logs if the matching log is not a CA:
```
CA certificate matching 'ca_trusted_fingerprint' found, adding it to 'certificate_authorities'
no CA certificate matching the fingerprint
```

With the changes here, on the same situation this will be logged:
```
Certificate matching 'ca_trusted_fingerprint' found, but is not a CA certificate
no CA certificate matching the fingerprint
```

## Why is it important?

To avoid confusion due to the contradictory logs.

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (there were tests covering this part).

### DIFF
--- a/transport/tlscommon/tls_config.go
+++ b/transport/tlscommon/tls_config.go
@@ -169,18 +169,23 @@ func trustRootCA(cfg *TLSConfig, peerCerts []*x509.Certificate) error {
 		// Compute digest for each certificate.
 		digest := sha256.Sum256(cert.Raw)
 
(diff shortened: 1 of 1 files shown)


pierrehilbert (approved):
LGTM

Issue #134, opened 2023-06-20
ReadPEMFile doesn't properly read PKCS#8 encrypted keys


jsoriano:
When reading a PKCS#8 encrypted key, `ReadPEMFile` returns the encrypted PEM block, even when it is expected to return it unencrypted. Any attempt to use the returned key results in `tls: failed to parse private key`, because it is still unencrypted.

`ReadPEMFile` relies on the deprecated `x509.IsEncryptedPEMBlock` and `x509.DecryptPEMBlock` deprecated functions for decryption. These functions rely on the `DEK-Info` header, included in encrypted private keys using PKCS#1 format and considered unsafe.

`ReadPEMFile` should decrypt keys in PKCS#8 format, or fail when it is not able to decrypt them.

jsoriano:
Closed by mistake.

PR #135, opened 2023-06-20
Parse encrypted private keys in PKCS#8 format

jsoriano:
## What does this PR do?

Add support to parse encrypted private keys in PKCS#8 format.

## Why is it important?

At least OpenSSL 3 uses PKCS#8 format by default. `ReadPEMFile` silently left private keys on this format encrypted, what produces problems later when trying to use them.

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Fixes https://github.com/elastic/elastic-agent-libs/issues/134.

### DIFF
--- a/transport/tlscommon/tls.go
+++ b/transport/tlscommon/tls.go
@@ -121,14 +121,14 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
 		case x509.IsEncryptedPEMBlock(block): //nolint: staticcheck // deprecated, we have to get rid of it
 			block, err := decryptPKCS1Key(*block, pass)
 			if err != nil {
-				log.Errorf("Dropping encrypted pem with private key '%v': %v", block.Type, err)
+				log.Errorf("Dropping encrypted pem block with private key, block type '%s': %s", block.Type, err)
 				continue
 			}
 			blocks = append(blocks, &block)
 		case block.Type == "ENCRYPTED PRIVATE KEY":
 			block, err := decryptPKCS8Key(*block, pass)
 			if err != nil {
-				log.Errorf("Dropping encrypted private key '%v': %v", block.Type, err)
+				log.Errorf("Dropping encrypted pem block with private key, block type '%s', could not decypt as PKCS8: %s", block.Type, err)
 				continue
 			}
 			blocks = append(blocks, &block)
--- a/transport/tlscommon/tls.go
+++ b/transport/tlscommon/tls.go
@@ -117,64 +117,24 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
 			break
 		}
 
-		if x509.IsEncryptedPEMBlock(block) { //nolint: staticcheck // deprecated, we have to get rid of it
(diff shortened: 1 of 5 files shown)


transport/tlscommon/tls.go
```diff
@@ -135,6 +139,41 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
...
+			case *ecdsa.PrivateKey:
+				block.Type = "ECDSA PRIVATE KEY"
+			default:
+				log.Errorf("Dropping encrypted pem of unknown type %T", key)
+				continue
+			}
+
+			buffer, err := x509.MarshalPKCS8PrivateKey(key)
+			if err != nil {
+				log.Errorf("Dropping encrypted pem of unknown type %T", key)
```
jsoriano, line 170: Fixing this error message and splitting this method in different functions...

transport/tlscommon/tls.go
```diff
@@ -113,29 +117,24 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
...
-				err = errors.New("no passphrase available")
-			} else {
-				// Note, decrypting pem might succeed even with wrong password, but
-				// only noise will be stored in buffer in this case.
-				buffer, err = x509.DecryptPEMBlock(block, pass) //nolint: staticcheck // deprecated, we have to get rid of it
+		switch {
+		case x509.IsEncryptedPEMBlock(block): //nolint: staticcheck // deprecated, we have to get rid of it
+			block, err := decryptPKCS1Key(*block, pass)
+			if err != nil {
+				log.Errorf("Dropping encrypted pem with private key '%v': %v", block.Type, err)
```
belimawr, line 124: I'm a bit confused by this error message, on a first glance I'd expect the private key to be logged as the first argument :scream: , but because the value is `block.Type` I believe it the message should better reflect what it means, something like that:
```suggestion
				log.Errorf("Dropping encrypted pem block with private key, block type '%s': %s", block.Type, err)
```

Also `block.Type` is a string, so `%s` is more appropriated there. For the error I also rather use `%s`, but that's totally up to you.

transport/tlscommon/tls.go
```diff
@@ -113,29 +117,24 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
...
+				continue
 			}
-
+			blocks = append(blocks, &block)
+		case block.Type == "ENCRYPTED PRIVATE KEY":
+			block, err := decryptPKCS8Key(*block, pass)
 			if err != nil {
-				log.Errorf("Dropping encrypted pem '%v' block read from %v. %+v",
-					block.Type, r, err)
+				log.Errorf("Dropping encrypted private key '%v': %v", block.Type, err)
```
belimawr, line 131: Same thing here, but also adding some context about what was happening:
```suggestion
				log.Errorf("Dropping encrypted pem block with private key, block type '%s', could not decypt as PKCS8: %s", block.Type, err)
```

belimawr (changes requested):

belimawr (approved):

PR #148, opened 2023-09-08
Add Elastic-Api-Version header to kibana client

pchila:
## What does this PR do?

Adds the `Elastic-Api-Version` header to all request to kibana Api as it's mandatory since merge of elastic/kibana#163570

This is the first step of a 2-step fox for issue elastic/elastic-agent#3362: after this PR is merged we need to bump the version and use the newer version in elastic-agent.

## Checklist

- ~~[ ] My code follows the style guidelines of this project~~
- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~
- ~~[ ] I have added an entry in `CHANGELOG.md`~~

## Author's Checklist

- [ ]

## Related issues

- Relates elastic/elastic-agent#3362

### DIFF
--- a/kibana/client.go
+++ b/kibana/client.go
@@ -191,14 +191,7 @@ func NewClientWithConfigDefault(config *ClientConfig, defaultPort int, binaryNam
 		binaryName = "Libbeat"
 	}
 	userAgent := useragent.UserAgent(binaryName, version, commit, buildtime)
-
-	const elasticAPIVersionHeaderKey = "Elastic-Api-Version"
-	const elasticAPIDefaultVersion = "2023-10-31"
-	rt, err := config.Transport.Client(
-		httpcommon.WithHeaderRoundTripper(
-			map[string]string{"User-Agent": userAgent, elasticAPIVersionHeaderKey: elasticAPIDefaultVersion},
-		),
-	)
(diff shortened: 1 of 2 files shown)


kibana/client.go
```diff
@@ -191,7 +191,14 @@ func NewClientWithConfigDefault(config *ClientConfig, defaultPort int, binaryNam
 		binaryName = "Libbeat"
 	}
 	userAgent := useragent.UserAgent(binaryName, version, commit, buildtime)
-	rt, err := config.Transport.Client(httpcommon.WithHeaderRoundTripper(map[string]string{"User-Agent": userAgent}))
+
+	const elasticAPIVersionHeaderKey = "Elastic-Api-Version"
+	const elasticAPIDefaultVersion = "2023-10-31"
```
leehinman, line 196: Could we just add "Elastic-Api-Version" to the `Headers` in `DefaultClientConfig()` function?  That way users can override if they need to for some reason and the defaults would be in the `DefaultClientConfig()` function where the other defaults of localhost & http are.

pchila: We could... even if now this is the only version that is supported :sweat_smile: 
Will change it

leehinman (approved):
LGTM

fearful-symmetry (approved):

PR #150, opened 2023-09-13
Add checks for serverless support 

fearful-symmetry:
## What does this PR do?
Part of https://github.com/elastic/ingest-dev/issues/2176 
and needed by https://github.com/elastic/elastic-agent/pull/3258

This adds a `KibanaIsServerless()` method to the kibana client, and also adds a header check to `ImportMultiPartFormFile`()`, as the underlying saved objects API requires a special header when used under serverless

## Why is it important?

Needed for serverless support.

## Checklist

### DIFF
--- a/kibana/client.go
+++ b/kibana/client.go
@@ -346,6 +346,19 @@ func (client *Client) readVersion() error {
 // IgnoreVersion was set when creating the client.
 func (client *Client) GetVersion() version.V { return client.Version }
 
+// KibanaIsServerless returns true if we're talking to a serverless instance.
+// Right now we don't have an API to tell us if we're running against serverless or not, so this actual implementation is something of a hack.
+// see https://github.com/elastic/kibana/pull/164850
+func (client *Client) KibanaIsServerless() (bool, error) {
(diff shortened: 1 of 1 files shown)


faec (approved):

kibana/client.go
```diff
@@ -346,6 +346,19 @@ func (client *Client) readVersion() error {
 // IgnoreVersion was set when creating the client.
 func (client *Client) GetVersion() version.V { return client.Version }
 
+// KibanaIsServerless returns true if we're talking to a serverless instance.
+// Right now we don't have an API to tell us if we're running against serverless or not, so this actual implementation is something of a hack.
+// see https://github.com/elastic/kibana/pull/164850
+func (client *Client) KibanaIsServerless() (bool, error) {
+	ret, _, err := client.Connection.Request("GET", "/api/saved_objects/_find", nil, nil, nil)
```
cmacknz, line 353: If Kibana changes this, which is entirely possible because saved objects are one of the major APIs, then this might give us the wrong result at some point. Hopefully we can catch that in the snapshot builds before release.

PR #169, opened 2023-12-19
Update TLS certificates in tests

belimawr:
## What does this PR do?

Some TLS certificates used in tests expired, this commit fixes it by generating the certificates and, if needed, calculating the fingerprint on each test.

This will prevent future CI failures and reduce the maintenance burden.

## Why is it important?

It fixes the broken tests

## Checklist

~~- [ ] I have added tests that prove my fix is effective or that my feature works~~

~~## Author's Checklist~~
~~## Related issues~~


blakerouse (approved):
Looks good. Thanks for fixing it where it won't need to be updated again.

transport/tlscommon/tls_config_test.go
```diff
@@ -678,30 +676,111 @@ func startTestServer(t *testing.T, serverAddr string, serverCerts []tls.Certific
...
+			data.keyUsage,
+			data.isCA,
+			certName,
+			data.dnsNames,
+			data.ips,
+			data.expired,
+		)
 		if err != nil {
-			t.Fatalf("parsing certificate %q: %+v", certname, err)
+			t.Fatal(err)
```
leehinman, line 751: can we please add back info about which cert failed?  If anything goes wrong I at least need to know which test case failed.  And I would probably want the key and cert as well.

transport/tlscommon/tls_config_test.go
```diff
@@ -678,30 +676,111 @@ func startTestServer(t *testing.T, serverAddr string, serverCerts []tls.Certific
...
 		}
-		certs[testcase] = testCert
+		certs[certName] = cert.Leaf
 	}
 
+	// If for any reason there is a need to debug
```
leehinman, line 756: If for some reason this fails in CI, we are going to want the cert/key pair to see what happened.  How about always dumping out the keys & certs, then deleting the dir if the test succeeds?

transport/tlscommon/ca_pinning_test.go
```diff
@@ -360,10 +360,19 @@ func genSignedCert(
...
 	dnsNames []string,
 	ips []net.IP,
+	expired bool,
 ) (tls.Certificate, error) {
 	if commonName == "" {
 		commonName = "You know, for search"
 	}
+
+	notBefore := time.Now()
+	notAfter := notBefore.Add(time.Hour)
```
leehinman, line 370: I would suggest more than 1 hour, just for insurance around daylight savings changes.  Yes, the golang time library should get that right, but I really don't want to debug some edge case if it doesn't.

leehinman (changes requested):

transport/tlscommon/tls_config_test.go
belimawr: Thanks Lee! I'l update them all together.

transport/tlscommon/tls_config_test.go
belimawr: Yeah, I first thought about it, but while writing this PR what helped me was to have a fixed set of certs that I could use over multiple test runs. So I ended up generating them once and re-using until I found out where the mismatch between cert, CA and fingerprint were.

Anyway, I see no issue in persisting them on failure.

transport/tlscommon/ca_pinning_test.go
belimawr: Good catch!

belimawr:
All suggestions implemented on 9d031ae

PR #171, opened 2023-12-20
[logp] Add typed loggers allowing log entries to go to different outputs

belimawr:
## What does this PR do?

This commit introduces a `typedLogger` that allows a log entry to be directed to different outputs when a configurable key/value are present in the log fields. 

This enables Beats (standalone and running under Elastic-Agent) to log event data to a separate log file.

## Why is it important?

It is required by:
- https://github.com/elastic/beats/pull/38767
- https://github.com/elastic/elastic-agent/pull/4549

## Checklist

- [x] Test `WithFileOrStderrOutput` to ensure it uses the correct config/level https://github.com/elastic/elastic-agent-libs/pull/171#discussion_r1462168918
- [x] Test the typedLoggerCore

## Related issues
- Required by https://github.com/elastic/beats/pull/37475
- Required by https://github.com/elastic/elastic-agent/pull/4129
- Required by https://github.com/elastic/beats/pull/38767
- Required by https://github.com/elastic/elastic-agent/pull/4549


logp/core.go
```diff
@@ -236,6 +236,22 @@ func makeEventLogOutput(cfg Config, enab zapcore.LevelEnabler) (zapcore.Core, er
 	return wrappedCore(core), nil
 }
 
+// WithFileOutput creates a new file output based on cfg and
+// replaces the previous one.
+func WithFileOutput(cfg Config) func(zapcore.Core) zapcore.Core {
```
cmacknz, line 241: Is it possible to write a test for this?

belimawr: Done on 76d680c

belimawr:
rebased onto `main`, force push.

logp/core.go
```diff
@@ -236,6 +236,22 @@ func makeEventLogOutput(cfg Config, enab zapcore.LevelEnabler) (zapcore.Core, er
...
+	out, err := makeFileOutput(cfg, zap.DebugLevel)
```
cmacknz, line 242: Is this hard coding the debug level?

belimawr: It was, thanks for catching that. It's fixed on 5cacb6c

cmacknz: Is there a way to update the test to have caught this?

logp/core.go
```diff
@@ -236,6 +236,34 @@ func makeEventLogOutput(cfg Config, enab zapcore.LevelEnabler) (zapcore.Core, er
...
+// is created. If both are true, the file output has preference.
+func WithFileOrStderrOutput(cfg Config) func(zapcore.Core) zapcore.Core {
+	var out zapcore.Core
+	var err error
+
+	switch {
+	case cfg.ToFiles:
+		out, err = makeFileOutput(cfg, cfg.Level.ZapLevel())
+	case cfg.ToStderr:
+		out, err = makeStderrOutput(cfg, cfg.Level.ZapLevel())
```
cmacknz, line 250: If both ToFiles and ToStderr are true we prefer ToFiles?

logp/core_test.go
```diff
@@ -212,3 +216,117 @@ func TestLoggingECSFields(t *testing.T) {
 		}
 	}
 }
+
+func TestWithFileOutput(t *testing.T) {
+	var tempDir1, tempDir2 string
+	// Because of the way logp and zap work, when the test finishes, the log
```
cmacknz, line 222: Add a reference to https://github.com/elastic/elastic-agent-libs/issues/179 so the next person to read this knows it can be removed.

logp/core_test.go
```diff
@@ -212,3 +216,117 @@ func TestLoggingECSFields(t *testing.T) {
...
+	if err := logger.Sync(); err != nil {
+		t.Fatalf("could not sync log file from fist logger: %s", err)
+	}
+
+	// Find the log file. The file name gets the date added, so we list the
+	// directory and ensure there is only one file there.
+	files, err := os.ReadDir(tempDir2)
+	if err != nil {
+		t.Fatalf("could not read temporary directory '%s': %s", tempDir2, err)
+	}
```
cmacknz, line 302: Shouldn't you also check that the first logger still works? This test would still pass if cloning the first logger broke it, or if writes to the second logger also went into the files of the first logger.

logp/core.go
belimawr: I believe be best approach is to test this function. It's a small one, so it should be simple

logp/core.go
belimawr: Yes, that's the indented behaviour, because it's sensitive data and when running in a container environment we log to stdout/stderr, I opted to have the file output having priority to avoid leaking event data by accident/missconfiguraiton.

logp/core_test.go
belimawr: Done d55bae1

logp/core_test.go
belimawr: Creating a new logger is not a new behaviour, hence I didn't think about looking at it's log  file/output. But I did ensure that if the first logger is used after the second is created and has its output replaced, the first one won't log to the same file as the second one.

Do you still believe we need to add another case? Like also reading/validating the log file for the first logger?

belimawr:
I updated the PR with the changes required for Elastic-Agent as well.

logp/core.go
belimawr: Done: 24878e6

belimawr:
I believe that's the final version.

belimawr:
I found a bug, I'm working on the fix and a test for it.

logp/core_test.go
belimawr: I added assertions to the original/default logger as well: b672556

logp/core.go
```diff
@@ -68,36 +68,33 @@ func Configure(cfg Config) error {
...
-	if cfg.toObserver {
+	if defaultLoggerCfg.toObserver {
 		sink, observedLogs = observer.New(level)
 	} else {
-		sink, err = createLogOutput(cfg, level)
+		sink, err = createLogOutput(defaultLoggerCfg, level)
 	}
 	if err != nil {
-		return fmt.Errorf("failed to build log output: %w", err)
+		return nil, level, nil, nil, fmt.Errorf("failed to build log output: %w", err)
```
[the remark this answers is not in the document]
belimawr: I don't believe so. It's an internal function and those values are not all tightly correlated.

logp/typedloggercore.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return nil
+}
+
+func (t *typedLoggerCore) Write(e zapcore.Entry, fields []zapcore.Field) error {
+	for _, f := range fields {
+		if f.Key == t.key {
+			if f.String == t.value {
+				return t.typedCore.Write(e, fields)
+			}
```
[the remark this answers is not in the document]
belimawr: > can we have multiple (f.Key == t.key) hits here?

No, `f.Key` need to be unique, otherwise they will end up as duplicated in the final JSON. While the code will accept it, there will be issues down the road.

Another reason to do not duplicate is is that `t.key/t.vlue` indicates the log type, so there can be only one set. A log entry cannot be `default` and `event` at the same time.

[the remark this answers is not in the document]
michalpristas: so we can break a loop once we hit f.Key==t.key and inner condition is not met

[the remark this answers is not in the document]
belimawr: Done in 1dca25a.

Thanks for spotting that!

belimawr:
Rebased onto `main`

faec (approved):

PR #176, opened 2024-01-08
add ServerConfig tags, export TLSClientAuth

michel-laterman:
## What does this PR do?

add ServerConfig tags, export TLSClientAuth

## Why is it important?

We want to enable mTLS args as part of elastic-agent's CLI. These args should be parsed and passed to fleet-server. YAML tags are needed to allow for serialization.

### DIFF
--- a/transport/tlscommon/server_config_test.go
+++ b/transport/tlscommon/server_config_test.go
@@ -38,7 +38,7 @@ func Test_ServerConfig_Serialization_ClientAuth(t *testing.T) {
 			},
 			CAs: []string{"/path/to/ca.crt"},
 		},
-		clientAuth: TLSClientAuthRequired,
+		clientAuth: TLSClientAuthNone, // NOTE the above config will be serialized with client_authentication: none
 	}, {
 		name: "no ca",
 		cfg: ServerConfig{
--- a/transport/tlscommon/server_config_test.go
+++ b/transport/tlscommon/server_config_test.go
@@ -24,11 +24,18 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// variables so we can use pointers in tests
+var (
+	required = TLSClientAuthRequired
+	optional = TLSClientAuthOptional
+	none     = TLSClientAuthNone
+)
+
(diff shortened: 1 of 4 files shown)


transport/tlscommon/server_config_test.go
```diff
@@ -0,0 +1,81 @@
...
+		name: "with ca and client auth none",
+		cfg: ServerConfig{
+			Certificate: CertificateConfig{
+				Certificate: "/path/to/cert.crt",
+				Key:         "/path/to/cert.key",
+			},
+			CAs:        []string{"/path/to/ca.crt"},
+			ClientAuth: TLSClientAuthNone,
+		},
+		clientAuth: TLSClientAuthNone, // FIXME test fails, it serializes to required
```
michel-laterman, line 61: This test is failing because of the `omitempty` tag on the ClientAuth attribute of the struct.
`TLSClientAuthNone` corresponds with a zero value, so it will not be serialized if it is set in the struct.

If we remove the `omitempty` tag, the "with ca" unit test will fail (none will serialize, and the `Unpack` method will not insert the required val (current default when a CA is specified).
I can try to change ClientAuth to be a pointer value instead, so there is a difference between no value, and a none (0) val

transport/tlscommon/server_config_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}{{
+		name: "with ca",
+		cfg: ServerConfig{
+			Certificate: CertificateConfig{
+				Certificate: "/path/to/cert.crt",
+				Key:         "/path/to/cert.key",
+			},
+			CAs: []string{"/path/to/ca.crt"},
+		},
+		clientAuth: TLSClientAuthNone, // NOTE the above config will be serialized with client_authentication: none
```
michel-laterman, line 41: I removed the `omitempty` from `ClientAuth`'s tags. This changed the behaviour when serializing a `ServerConfig` with `CAs` but no `ClientAuth` (Unpack no longer injects `required` on deserialization)

blakerouse (approved):
Looks good.

Thanks for all the additional testing.

PR #183, opened 2024-02-12
Add optimized ReadAll for http responses

rdner:
## What does this PR do?

1. When the body content length is known it's much faster to pre-allocated the entire buffer once.
2. When the length is unknown using `bytes.NewBuffer` + `io.Copy` is much faster.

Also, added benchmarks to prove the difference.

## Why is it important?

When we start using this function in main products it will significantly improve the performance.

## Checklist

## Benchmark

```
goos: darwin
goarch: arm64
pkg: github.com/elastic/elastic-agent-libs/transport/httpcommon
BenchmarkReadAll/size:_100/unknown_length/io.ReadAll-10                       	    1000	      1232 ns/op	     576 B/op	       3 allocs/op
BenchmarkReadAll/size:_100/unknown_length/bytes.Buffer+io.Copy-10         	    1000	      1219 ns/op	     224 B/op	       4 allocs/op

BenchmarkReadAll/size:_100/known_length/io.ReadAll-10                     	    1000	      1275 ns/op	     576 B/op	       3 allocs/op
BenchmarkReadAll/size:_100/known_length/bytes.Buffer+io.Copy-10           	    1000	      1206 ns/op	     176 B/op	       3 allocs/op

BenchmarkReadAll/size:_102400/unknown_length/io.ReadAll-10                	    1000	     60858 ns/op	  514405 B/op	      24 allocs/op
BenchmarkReadAll/size:_102400/unknown_length/bytes.Buffer+io.Copy-10      	    1000	      9833 ns/op	  106640 B/op	       8 allocs/op

BenchmarkReadAll/size:_102400/known_length/io.ReadAll-10                  	    1000	     44423 ns/op	  514401 B/op	      24 allocs/op
BenchmarkReadAll/size:_102400/known_length/bytes.Buffer+io.Copy-10        	    1000	      9229 ns/op	  106592 B/op	       7 allocs/op

BenchmarkReadAll/size:_1048576/unknown_length/io.ReadAll-10               	    1000	    474003 ns/op	 5241202 B/op	      33 allocs/op
BenchmarkReadAll/size:_1048576/unknown_length/bytes.Buffer+io.Copy-10     	    1000	     81912 ns/op	 1048722 B/op	       8 allocs/op

BenchmarkReadAll/size:_1048576/known_length/io.ReadAll-10                 	    1000	    457693 ns/op	 5241196 B/op	      33 allocs/op
BenchmarkReadAll/size:_1048576/known_length/bytes.Buffer+io.Copy-10       	    1000	     80061 ns/op	 1048674 B/op	       7 allocs/op
PASS
ok  	github.com/elastic/elastic-agent-libs/transport/httpcommon	1.543s
```

## Related issues

- Relates https://github.com/elastic/beats/issues/36151

### DIFF
--- a/transport/httpcommon/httpcommon.go
+++ b/transport/httpcommon/httpcommon.go
@@ -18,6 +18,10 @@
 package httpcommon
 
 import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


belimawr (approved):

amitkanfer:
❤️

cmacknz:
https://github.com/golang/go/issues/50774 appears to be the upstream Go issue tracking the inefficiency of `io.ReadAll` here. Linking it for future reference.

Quoting the explanation of what causes this:

> The underlying reason is that ReadAll() uses append to resize its buffer, which goes up by a factor of 1.25 (slightly more in latest Go, but asymptotically the same) each time.
> 
> One might measure these alternatives on various axes:
> 
> number of bytes allocated and freed
> max size of heap required to read a given size
> amount of copying while resizing
> percentage waste in returned buffer
>
> The current implementation is better on the last one, and much worse on all the others.
>
> append is designed to minimize space wasted at the end of the buffer

PR #187, opened 2024-02-25
Optimize clone method

Dhruvit96:
## What does this PR do?

This PR improves the performance of the Clone method for the M type in `mapstr` package.
It uses an inbuilt method for cloning maps which is more efficient.
ref: https://github.com/golang/go/issues/58740

## Why is it important?

This change is important as the Clone method is used before applying processors in the beats repo each time.
This small optimization will improve the performance of the Beats pipeline by a substantial margin.

## Checklist


Dhruvit96:
I've signed CLA.

faec:
Thanks for pointing this out! I'm setting up some benchmarks to see how this compares to the current version.

dhruvit-crest:
> Thanks for pointing this out! I'm setting up some benchmarks to see how this compares to the current version.

Thank you please let me know if I can help any way.

faec:
I've done some benchmarks and the results seem odd. All the CPU numbers are neutral-leaning-higher with the new code. I've taken profiles and confirmed that everything is using the new optimized stdlib, and it does save ~10% of copying time (~1% of overall cpu in the benchmarks I ran), but then it loses that time back in increased GC costs.

The reported CPU reduction in the stdlib change is more like 85%, so this is a pretty big mismatch. I'm going to do some more testing to try and track this down.

One thing to add to the PR, though: the baseline version of `Clone` always returns a non-nil result, even when called on `nil`, whereas your version returns nil, and that causes crashes since Filebeat assumes the return value is non-nil. (It maybe shouldn't, when cloning a nil pointer, but that's the current behavior.) To avoid any surprise crashes can you add a nil check on the input to explicitly return a non-nil empty map?

Dhruvit96:
Thanks @faec for looking into it, I'm curious about what tests you have run. So far I have only used Go's framework for benchmarking and pprof library. If you could point to articles or tools that you're using or referring to, it would be a great help to my learning.

Also thanks for pointing out behavior mismatch. I'll add a nil check to maintain the current behavior.

mauri870:
Closing this because our clone implementation has changed since this was proposed, so it is no longer relevant.

Some considerations:

- [maps.Clone](https://cs.opensource.google/go/go/+/go1.23.6:src/maps/maps.go;l=50) performs a shallow copy, meaning that modifying the cloned map will also modify the original map. Our doc comment for Clone does not explicitly state this (perhaps it should), but we rely on it performing a deep copy, where the resulting map holds no references to the original.
- With the [recent changes](https://github.com/elastic/elastic-agent-libs/blob/e11ce6b/mapstr/mapstr.go#L164), Clone is as fast as possible, essentially using a for loop with map assignments to a new map. Go provides [maps.Copy](https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/maps/maps.go;l=62), which works similarly. Therefore, iterating over the map and copying its fields into a new map remains the most efficient approach.

PR #189, opened 2024-03-11
Add testing/certutil/ package for TLS test support

AndersonQ:
## What does this PR do?

Adds the `testing/certutil/` package providing on-the-fly generation of TLS certificate and root CA and a CLI to generate them as well.

## Why is it important?

The Elastic Agent and Bats need to configure TLS for testing. This provide an easy way to generate the needed certificates without duplicating code in the Elastic Agent and Beats repo

## Checklist

## Related issues

- Relates elastic/elastic-agent/issues/2247


leehinman:
I am strongly in favor of generating the CA & certs as needed for testing.  But do we need to write our own utility?

https://github.com/cloudflare/cfssl

Already does all of this, is go so we can embed in our tests, and has been around a while so most of the subtle X509 considerations have been addressed.

Downside is that we don't build up as much X509 knowledge in our team.  Thoughts?

AndersonQ:
> I am strongly in favor of generating the CA & certs as needed for testing. But do we need to write our own utility?
> 
> https://github.com/cloudflare/cfssl
> 
> Already does all of this, is go so we can embed in our tests, and has been around a while so most of the subtle X509 considerations have been addressed.
> 
> Downside is that we don't build up as much X509 knowledge in our team. Thoughts?

I was having a look at it. It feels overkill and I'd bet getting used to use it would not be much different than using what this PR adds. Also as you pointed, by building it ourselves we get some more knowledge in the team about how all the moving pieces get together.

Lastly, it's done already, the beats use a similar code that can, and should, be replaced by this new package. Just the work to replace this PR and instead use `github.com/cloudflare/cfssl` would not be worth.

leehinman (changes requested):

AndersonQ:
as discussed directly with @leehinman, the child certificate generation will be improved on another PR and use the `Certificate Request`

leehinman (approved):

PR #190, opened 2024-03-18
tlscommon: fix EC key decryption

AndersonQ:
## What does this PR do?

Fix how encrypted TLS certificate keys are handled. `x509.DecryptPEMBlock` returns a PEM block, and the code would assume it was not in PEM format and was re-encoding it into PEM, which would generate an invalid key.

## Why is it important?

Without the fix, it breaks when parsing a certificate key 

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/2247
- Relates https://github.com/elastic/elastic-agent/issues/2248

### DIFF
--- a/transport/tlscommon/tls.go
+++ b/transport/tlscommon/tls.go
@@ -93,7 +93,6 @@ func LoadCertificate(config *CertificateConfig) (*tls.Certificate, error) {
 // and decrypt it with the provided password and  return the raw content.
 func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
 	pass := []byte(passphrase)
-	var blocks []*pem.Block
 
 	r, err := NewPEMReader(s)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmacknz:
Needs a test

transport/tlscommon/tls.go
```diff
@@ -106,49 +105,49 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
...
-			blocks = append(blocks, &block)
+			// not in PEM format, encode it.
+			err = pem.Encode(buffer, &block)
+			if err != nil {
+				log.Errorf("Dropping encrypted pem block with private key, block type '%s' could not PEM encode it: %s", block.Type, err)
+				continue
+			}
 		default:
-			blocks = append(blocks, block)
+			// not encrypted and not in PEM format, encode it.
```
leehinman, line 138: Is this function only used for reading PEM encoded private keys?  The implementation suggests that, but the function name and signature don't.  If this were to read a PEM encoded certificate, you don't want to re-encode.

As it stands, I think the default case is trying to catch if they actually supplied a DER encoded private key.  If that is the case it is probably worth checking to see if `ParsePKIXPublicKey` works and to log at the warn level that DER format was provided but PEM format was expected.

leehinman (changes requested):
needs tests

AndersonQ:
`x509.DecryptPEMBlock` should receive a DER encoded key, not a PEM encoded. That way the key it returns will always be in DER format and we need to encode it as PEM. Thus the code is actually correct

PR #196, opened 2024-04-10
Add support for repacking and merging tls.Config structs

michel-laterman:
## What does this PR do?

Add support for the Unpack operation for tls.Config attributes that are specified as strings but stored as other values. The existing Unpack operations are limited to string only and these will cause issues when trying to merge previously unpacked structs.

## Why is it important?

We are tying to get fleet-server instances running in agent mode to reload its output block using values retrieved from the policy. See [the pr](https://github.com/elastic/fleet-server/pull/3411)
Merging the changed output value into the existing config is needed to support this.
However we are running into an error string `verification mode must be an identifier accessing 'inputs.0.server.ssl.verification_mode' accessing 'inputs.0.server.ssl'` that occurs because we are working with structs that have been already unpacked.
The types that cause issue are those that are serialized as a string but represented in the struct as a numeric value.

## Checklist

- ~~I have commented my code, particularly in hard-to-understand areas~~

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/2784

### DIFF
--- a/transport/tlscommon/types.go
+++ b/transport/tlscommon/types.go
@@ -159,28 +159,37 @@ func (m TLSVerificationMode) MarshalText() ([]byte, error) {
 	return nil, fmt.Errorf("could not marshal '%+v' to text", m)
 }
 
-// Unpack unpacks the string into constants.
+// Unpack unpacks the input into a TLSVerificationMode.
 func (m *TLSVerificationMode) Unpack(in interface{}) error {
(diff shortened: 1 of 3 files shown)


jlind23:
@pierrehilbert this is pending on the Elastic Agent team, can we get some traction please?

faec (approved):

PR #198, opened 2024-04-16
Add both uint64 and int64 to tlscommon types unpack methods

michel-laterman:
## What does this PR do?

Add `int64` support to unpack methods which only supported `uint64`

## Why is it important?

We see the error message
```
tls renegotation support is an unknown type: int64 accessing 'output.elasticsearch.ssl.renegotiation'
```
in our snapshots.

## Checklist

## Related issues

-

### DIFF
--- a/transport/tlscommon/server_config_test.go
+++ b/transport/tlscommon/server_config_test.go
@@ -173,3 +173,80 @@ func Test_ServerConfig_Repack(t *testing.T) {
 		})
 	}
 }
+
+func Test_ServerConfig_RepackJSON(t *testing.T) {
+	tests := []struct {
+		name string
+		json string
+		auth *TLSClientAuth
+	}{{
+		name: "with client auth",
+		json: `{
+    "enabled": true,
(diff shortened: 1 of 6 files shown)


cmacknz:
Needs tests

cmacknz:
(no tests is why we didn't catch this initially)

michel-laterman:
Not sure how the fleet-server is reading/unserializing the policies from es; the `Test*Repack` tests added previously didn't trigger this

transport/tlscommon/server_config_test.go
```diff
@@ -173,3 +173,80 @@ func Test_ServerConfig_Repack(t *testing.T) {
 		})
 	}
 }
+
+func Test_ServerConfig_RepackJSON(t *testing.T) {
```
michel-laterman, line 177: I'm not sure if we encounter json serialization in the error we observed but i've added tests in case that is the issue.
I've also added Unpack unit tests to all the types

cmacknz (approved):

PR #204, opened 2024-05-13
Disable logging when all outputs are disabled

belimawr:
## What does this PR do?

Disable logging when all outputs are disabled

## Why is it important?

Even when all log outputs were disabled in the configuration, we were still instantiating a file output. This is not the correct behaviour. If all log outputs are disabled, there should be no log output.

## Checklist

~~## Author's Checklist~~

## Related issues
- https://github.com/elastic/beats/pull/38767

### DIFF
--- a/logp/core.go
+++ b/logp/core.go
@@ -211,7 +211,7 @@ func createLogOutput(cfg Config, enab zapcore.LevelEnabler) (zapcore.Core, error
 	case SystemdEnvironment, ContainerEnvironment:
 		return makeStderrOutput(cfg, enab)
 	case MacOSServiceEnvironment, WindowsServiceEnvironment:
-		fallthrough
+		return makeFileOutput(cfg, enab)
 	default:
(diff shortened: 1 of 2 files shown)


logp/core.go
```diff
@@ -213,7 +213,7 @@ func createLogOutput(cfg Config, enab zapcore.LevelEnabler) (zapcore.Core, error
 	case MacOSServiceEnvironment, WindowsServiceEnvironment:
 		fallthrough
 	default:
-		return makeFileOutput(cfg, enab)
```
rdner, line 216: This does not disable logging when it's systemd or container environment, is it intended?

belimawr: Not, but it was disabling logging when it the environment was `MacOSServiceEnvironment` or `MacOSServiceEnvironment` because of the `fallthrough. I fixed it in 7b4c5e5

rdner (approved):

PR #205, opened 2024-05-17
Add Close method to logp.Logger

belimawr:
## What does this PR do?

Add a close method to logp.Logger responsible for closing any underlying core/output. All of our implementations of zapcore.Core that use an output that can be closed now implement io.Closer.

This allows for closing any underlying files used by any output, which can prevent issues on Windows.

It also removes the workaround for testing the loggers on Windows.

## Why is it important?

It allows us to correctly close any underlying file descriptor used by our loggers. This allows us to make sure no log data is lost as well as prevent issues on Windows where folders containing active log files cannot be removed by the Elastic-Agent

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works

~~## Author's Checklist~~

## Related issues
- Closes https://github.com/elastic/elastic-agent-libs/issues/179

PR #208, opened 2024-05-29
[logp] Default logging config logs to file

belimawr:
## What does this PR do?

The default log configuration must log to files. A test is added to ensure the default logging configuration logs to files as well as some other default fields are correctly set.

## Why is it important?

It fixes the default behaviour that was broken by https://github.com/elastic/elastic-agent-libs/pull/205

## Checklist

## Author's Checklist
- [ ] Test Beats with this change
- [ ] Test Elastic-Agent with this change

~~## Related issues~~

### DIFF
--- a/logp/defaults_test.go
+++ b/logp/defaults_test.go
@@ -1,0 +1,111 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package logp_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/elastic/elastic-agent-libs/logp"
+)
+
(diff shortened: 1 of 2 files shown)


logp/defaults_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	// Get the logger and log anything
+	logger := logp.L()
+	defer logger.Close()
+
+	logger.Info("foo")
+	_, fileName, lineNum, _ := runtime.Caller(0)
+	lineNum-- // We want the line number from the log
+	fileName = filepath.Base(fileName)
```
cmacknz, line 55: Why do we need this here? This is really a test of what the `logger.Info` calls does isn't it? 

The point of this test is to prove that the default configuration logs to files, not to also test exact format of the logs themselves? Is that correct?

I think I am mostly concerned that this test succeeding currently depends on the exact placement of these lines in the file. It seems unnecessary to maintain this to verify what you need to.

belimawr: > The point of this test is to prove that the default configuration logs to files, not to also test exact format of the logs themselves? Is that correct?

Yes, but as I wrote the test I thought it would be nice to ensure some other fields we expect are there as well. The file and line number are essential exactly which bit of the code is producing the log entry. They're the first thing I look at when I'm dealing with a log entry I'm not used to see.

I don't mind removing this assertion, however I believe we should have a test that ensures our expected default behaviour, which includes some essential fields.

> I think I am mostly concerned that this test succeeding currently depends on the exact placement of these lines in the file. It seems unnecessary to maintain this to verify what you need to.

I get that, hence I'm reading  them from the runtime instead of just hard-coding them. As long as L52-54 are always together, no matter where in the file the tests will succeed.

I prefer adding a comment clarifying what those lines are doing and stating they must stay together rather than removing this assertion.

However, if you insist, I'm not against removing them for now and later writing a more comprehensive test with our expectations for a default log entry.

cmacknz: I don't hate it, I just wanted to sanity check how necessary this was. The root of my concern is that this could be fragile, and also if this functionality is already tested as a core part of the upstream zap logger it uses underneath. 

It's cheap enough to have here, it is not worth going on a separate adventure to perfectly test every aspect of the logger right now.

cmacknz (approved):

PR #216, opened 2024-07-04
Fix download serialization to JSON to be a float in bps instead of a string with unit

nchaulet:
## Description 

Related to https://github.com/elastic/ingest-dev/issues/3446

Fix DownloadRate serialization to JSON to be a float in bps instead of a string with unit.

as serializing `Inf` to json is not possible, I fallback to the max float value, I think it's probably better than crashing for that.


upgrade/details/details.go
```diff
@@ -33,32 +29,10 @@ type DownloadRate float64
 func (dr *DownloadRate) MarshalJSON() ([]byte, error) {
 	downloadRateBytesPerSecond := float64(*dr)
 	if math.IsInf(downloadRateBytesPerSecond, 0) {
-		return json.Marshal("+Inf bps")
+		return json.Marshal(math.MaxFloat64)
```
belimawr, line 32: I fear that this could be misleading, especially if we use this value to get metrics or averages in ES/Kibana...

I get that having the unit is causing problems, but how often is this value actually `Inf`? Is it because the download rate is too slow or too fast?

I believe a custom marshal/unmarshal here and on Fleet-Server could more accurately handle those situations.

belimawr (changes requested):

upgrade/details/details.go
nchaulet: @belimawr actually I am not sure we ever encounter the Inf issue ever, as with the max float 64 value we could support `1.798e+284YBps`,

belimawr: Then I believe the best is to encode as `+Inf/-Inf` and deal with that on the Fleet side as well. Or even not send this value/piece of data as we know it is due to some issue.

The key thing on my mind is that if we show this to our users, it needs to be accurate, or clearly state we can't get this metric for some reason.

ycombinator: Yes, IIRC, we have occasionally seen `+Inf` when the download happens really fast, almost instantaneously.  But, stepping back and taking a look at the end-goal here, it is to let the user know how fast a download is proceeding.  For really fast downloads, a user is not going to care if we report that as `+Inf bps` (which, of course, has proven problematic being a string) or some really large number like `math.MaxFloat64`.  In some ways, the latter is a bit more realistic even.  So I'm +1 for going with `math.MaxFloat64` here as the fix.

belimawr: If you two agree with `math.MaxFloat64` I'm not opposed to it. I'd just make sure we get the correct signal to differentiate between `+Inf` and `-Inf`.

belimawr (approved):

PR #218, opened 2024-07-08
configure logging from a zap.Core

leehinman:
## What does this PR do?

Adds new `ConfigureWithCore` to the logp package.  This allows you to use an existing `zapcore.Core` instead of creating a new one.  This is useful if you need to use logp within a project that has already setup a `zapcore.Core` on it's own.

## Why is it important?

Allows projects that use logp for logging to use an existing `zapcore.Core`

## Checklist

## Author's Checklist

- [ ]

## Related issues

- Relates elastic/beats#40110

### DIFF
--- a/logp/core.go
+++ b/logp/core.go
@@ -151,6 +151,56 @@ func ConfigureWithOutputs(defaultLoggerCfg Config, outputs ...zapcore.Core) erro
 	return nil
 }
 
+// ConfigureWithCore configures the global logger to use an output created
+// from `defaultLoggerCfg` and all the output passed by `output`.
+func ConfigureWithCore(defaultLoggerCfg Config, core zapcore.Core) error {
+	var (
+		sink         zapcore.Core
+		level        zap.AtomicLevel
+	)
+
+	level = zap.NewAtomicLevelAt(defaultLoggerCfg.Level.ZapLevel())
+	sink = wrappedCore(core)
+	
+	// Default logger is always discard, debug level below will
+	// possibly re-enable it.
+	golog.SetOutput(io.Discard)
+
+	// Enabled selectors when debug is enabled.
+	selectors := make(map[string]struct{}, len(defaultLoggerCfg.Selectors))
+	if defaultLoggerCfg.Level.Enabled(DebugLevel) && len(defaultLoggerCfg.Selectors) > 0 {
+		for _, sel := range defaultLoggerCfg.Selectors {
+			selectors[strings.TrimSpace(sel)] = struct{}{}
+		}
+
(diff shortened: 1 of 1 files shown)


logp/core.go
```diff
@@ -151,6 +151,56 @@ func ConfigureWithOutputs(defaultLoggerCfg Config, outputs ...zapcore.Core) erro
 	return nil
 }
 
+// ConfigureWithCore configures the global logger to use an output created
+// from `defaultLoggerCfg` and all the output passed by `output`.
+func ConfigureWithCore(defaultLoggerCfg Config, core zapcore.Core) error {
```
belimawr, line 156: @leehinman, I'm confused here, how is this function conceptually different from `ConfigureWithOutputs`?

Even the GoDoc for it is pretty much the same :P

cmacknz:
Is there a unit test that could be written for this function?

logp/core.go
leehinman: observedLogs and the default logger output.  But yes, these are very similar functions.  I'm fine with saying we should refactor.  The reason for opening the PR was to get you to look at it :-)

leehinman:
Converted to draft PR due to PTO and I think Tiago and I need to talk through if we can just use `ConfigureWithOutputs`.

logp/core.go
```diff
@@ -151,6 +151,48 @@ func ConfigureWithOutputs(defaultLoggerCfg Config, outputs ...zapcore.Core) erro
 	return nil
 }
 
+// ConfigureWithCore configures the global logger to use the passed in
+// core.  It is assumed that an output has already been defined with
```
AndersonQ, line 155: [NIT]
Is there any reason to use double spaces after the period?

AndersonQ (approved):
LGTM, but you may want to see if Tiago still has any comments

logp/core.go
mauri870: I recall seeing this in some other PRs from Lee, maybe it is an editor configuration on his side?

leehinman: This is just for context, since you both seem interested.  I'm not getting into a flame war about white space formatting in comments.

It has to do with being old enough that I learned to type on an actual [typewriter](https://en.wikipedia.org/wiki/IBM_Selectric).  2 spaces between sentences were **required**.  To see why they were required take a look at a full sheet of text in Pica 10pt (Courier is pretty close), and make sure justification is disabled.  Without the 2 spaces it is very hard to scan text to see where one sentence ends and one begins.  And since you are using your eyes to "search", this is a very desirable feature.

PR #220, opened 2024-08-01
Add periodic/Doer for frequency-limited action execution

AndersonQ:
## What does this PR do?

Add `periodic/Doer` for frequency-limited action execution

## Why is it important?

The Doer limits the execution of an action to at most once within a specified period. It's designed for managing frequent events by providing a summary with the number of occurrences within the given period. If no event happen, the action is not executed.

## Checklist

## Related issues

- Relates https://github.com/elastic/beats/issues/40157

### DIFF
--- a/periodic/ratelimited.go
+++ b/periodic/ratelimited.go
@@ -1,0 +1,120 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package periodic
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Doer limits an action to be executed at most once within a specified period.
+// It is intended for managing events that occur frequently, but instead of an
+// action being taken for every event, the action should be executed at most
+// once within a given period of time.
+//
+// Doer takes a function to execute, doFn, which is called every time
+// the specified period has elapsed with the number of events and the period.
+type Doer struct {
+	count atomic.Uint64
+
+	period time.Duration
+
+	// doFn is called for executing the action every period if at least one
+	// event happened. It receives the count of events and the period.
+	doFn     func(count uint64, d time.Duration)
+	lastDone time.Time
+	done     chan struct{}
+
+	// nowFn is used to acquire the current time instead of time.Now so it can
+	// be mocked for tests.
+	nowFn func() time.Time
+	// newTickerFn is used to acquire a *time.Ticker instead of time.NewTicker
+	// so it can be mocked for tests.
+	newTickerFn func(duration time.Duration) *time.Ticker
+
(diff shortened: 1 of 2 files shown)


logp/ratelimited.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// RateLimitedLogger is a logger that limits log messages to once within a
+// specified period.
+// It is intended for logging events that occur frequently, providing a summary
```
cmacknz, line 11: Why not use https://pkg.go.dev/golang.org/x/time/rate or even https://github.com/elastic/beats/tree/main/libbeat/processors/ratelimit

cmacknz: There is a custom implementation of the token bucket algorithm in https://github.com/elastic/beats/blob/main/libbeat/processors/ratelimit/token_bucket.go but I'd honestly prefer we replace that the one in x/time/rate as well.

AndersonQ: well apart from the fact I missed there is a `[Every](https://pkg.go.dev/golang.org/x/time/rate#Every)`, the work to accumulate the number of events to then log the summary would be the same. Also using the ticker it's a rather straight forward implementation. Thus I don't find it necessary to use https://pkg.go.dev/golang.org/x/time/rate

pierrehilbert:
@AndersonQ shouldn't we also change the PR title?

AndersonQ:
> @AndersonQ shouldn't we also change the PR title?

I did, but didn't click on 'save' ahah. Now it's fixed, thanks

periodic/ratelimited.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (r *Doer) AddN(n uint64) {
+	r.count.Add(n)
+}
+
+func (r *Doer) Start() {
+	if r.started.Load() {
+		return
+	}
```
rdner, line 79: This is not thread-safe, you should use the CAS operation instead https://pkg.go.dev/sync/atomic@go1.22.6#Bool.CompareAndSwap

periodic/ratelimited.go
```diff
@@ -0,0 +1,120 @@
...
+
+	r.done = make(chan struct{})
+	r.started.Store(true)
```
rdner, line 82: This should be removed because of the requested change above.

periodic/ratelimited.go
```diff
@@ -0,0 +1,120 @@
...
+		defer r.wg.Done()
+
+		defer r.ticker.Stop()
+
+		for {
+			select {
+			case <-r.ticker.C:
+				r.do()
+			case <-r.done:
+				r.do()
```
rdner, line 97: Are you sure we should run `do` here when we have the `done` signal?

periodic/ratelimited.go
```diff
@@ -0,0 +1,120 @@
...
+				return
+			}
+		}
+	}()
+}
+
+func (r *Doer) Stop() {
+	if !r.started.Load() {
+		return
+	}
```
rdner, line 107: This should be a CAS operation too. Atomic works only when you read and change the value in one operation.

periodic/ratelimited.go
```diff
@@ -0,0 +1,120 @@
...
+
+	close(r.done)
+	r.wg.Wait()
+	r.started.Store(false)
```
rdner, line 111: This should be removed due to the requested change above.

rdner (changes requested):
Not thread-safe.

periodic/ratelimited.go
AndersonQ: It's thread safe but I get your point, the value could change before `Store` is called.

periodic/ratelimited.go
AndersonQ: yes, otherwise we'll lose the events that happened between the last tick and `Stop` being called. Worst case it ticks and immediately after `Stop` is called, with no event happening in-between. In that case, do will be a no-op

rdner (approved):

PR #224, opened 2024-08-28
refactor: update unix build tags and add plan9

kruskall:
## What does this PR do?

resolve obscure compile errors when trying to compile on different goos.
Use unix build tags and add plan9.

We can't implement uid/gid logic without breaks as plan9 uses strings while the api requires int.
Use plan9.Dup to redirect stderr.

See https://man2.aiju.de/2/dup

On plan9, 
- plan9.Dup(oldfd, newfd) == unix.Dup2(oldfd, newfd)
- plan9.Dup(fd, -1) == unix.Dup(fd)

## Why is it important?

- `!windows` build tags is not true and causes misleading compile errors because it implies the file can be used in any goos other than windows while it's meant to be used for `unix`
- `_other` has the same problem

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/file/fileinfo_unix.go
+++ b/file/fileinfo_unix.go
@@ -22,8 +22,7 @@ package file
 import (
 	"errors"
 	"os"
-
-	"golang.org/x/sys/unix"
+	"syscall"
 )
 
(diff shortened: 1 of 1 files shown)


VihasMakwana (approved):
LGTM

PR #233, opened 2024-10-01
fix(logp): TestConfigureWithCore fails with Go 1.24

mauri870:
## What does this PR do?

The cmd/vet in Go 1.24 reports printf calls with non-const format and no args, causing one test in logp to fail.

See https://github.com/golang/go/issues/60529.

```
$ go install golang.org/dl/gotip@latest
$ gotip download
$ gotip test ./logp -count=1
logp/core_test.go:714:7: non-constant format string in call to github.com/elastic/elastic-agent-libs/logp.Info
FAIL    github.com/elastic/elastic-agent-libs/logp [build failed]
```

## Checklist

### DIFF
--- a/logp/core_test.go
+++ b/logp/core_test.go
@@ -711,7 +711,7 @@ func TestConfigureWithCore(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unexpected err: %s", err)
 	}
-	Info(testMsg)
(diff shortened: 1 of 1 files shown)


leehinman (approved):
Thanks for fixing.

mauri870:
> Thanks for fixing.

Appreciate the review, Lee. I'll open follow ups on elastic-agent and beats as needed since they also have similar failures.

PR #236, opened 2024-10-07
logp: don't write to files by default if running in a container environment

mauri870:
## What does this PR do?

The PR https://github.com/elastic/elastic-agent-libs/pull/208 introduced a bug that made  the `-environment container|systemd` flag stop writting to stderr on the default logger. This PR restores the old behavior as well as adding a test that observes that logs are properly written to stderr.

## Checklist

## Related issues

- Fixes https://github.com/elastic/beats/issues/41118

### DIFF
--- a/testing/certutil/certutil.go
+++ b/testing/certutil/certutil.go
@@ -71,14 +71,45 @@ func NewRSARootCA() (crypto.PrivateKey, *x509.Certificate, Pair, error) {
 	return rootKey, cert, pair, err
 }
 
-// GenerateChildCert generates a x509 Certificate as a child of caCert and
-// returns the following:
-// - the certificate in PEM format as a byte slice
-// - the private key in PEM format as a byte slice
+// GenerateChildCert generates a ECDSA (P-384) x509 Certificate as a child of
+// caCert and returns the following:
 // - the certificate and private key as a tls.Certificate
+// - a Pair with the certificate and its key im PEM format
 //
 // If any error occurs during the generation process, a non-nil error is returned.
-func GenerateChildCert(
+func GenerateChildCert(name string, ips []net.IP, caPrivKey crypto.PrivateKey, caCert *x509.Certificate) (*tls.Certificate, Pair, error) {
+	priv, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, Pair{}, fmt.Errorf("could not create RSA private key: %w", err)
+	}
+
+	cert, childPair, err :=
+		GenerateGenericChildCert(
+			name,
+			ips,
+			priv,
+			&priv.PublicKey,
+			caPrivKey,
+			caCert)
+	if err != nil {
+		return nil, Pair{}, fmt.Errorf(
+			"could not generate child TLS certificate CA: %w", err)
+	}
+
+	return cert, childPair, nil
+}
+
+// GenerateGenericChildCert generates a x509 Certificate using priv and pub
+// as the certificate's private and public keys and as a child of caCert.
+// Use this function if you need fine control over keys or ips and certificate name,
+// otherwise prefer GenerateChildCert or NewRootAndChildCerts/NewRSARootAndChildCerts
+//
+// It returns the following:
+// - the certificate and private key as a tls.Certificate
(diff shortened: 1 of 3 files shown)


logp/config.go
```diff
@@ -70,9 +70,16 @@ const (
 // DefaultConfig returns the default config options for a given environment the
 // Beat is supposed to be run within.
 func DefaultConfig(environment Environment) Config {
+	toFiles := true
+
+	// If running in a container environment, don't write to files by default.
+	if environment == ContainerEnvironment {
```
cmacknz, line 76: The elastic-agent container always logs to files in the file system by default, even if logging to stdout+stderr is disabled. These are the logs that are included in diagnostics from the container.

Let's make sure we don't break agent diagnostics when this merges into elastic-agent.

mauri870: ACK. This is reverting to the previous behavior. When we silently made this change earlier, it didn’t break elastic-agent, nor was it broken before the change. I don’t believe it's affected at all, but I could be wrong.

mauri870: So looking at changes in elastic-agent, there is only two occurrences of [`logp.DefaultConfig`](https://github.com/search?q=repo%3Aelastic%2Felastic-agent%20logp.DefaultEnvironment&type=code) and both use `logp.DefaultEnvironment` which always gets `ToFiles = true`.

belimawr (commented):
I was thinking a bit more about it and I believe the problem is not just because we always set `toFiles` to `true`, it's also because `DefaultConfig` sets an output, so when `logp.createLogOutput` https://github.com/elastic/elastic-agent-libs/blob/196dace11604a4403f4466812a150ff993c0261c/logp/core.go#L250-L272
gets called, the environment is not taken into consideration because an output has already been explicitly chosen.

I believe the systemd case (Filebeat running as a service) is also affected. Installing from the `deb` should trigger this case, I'm not sure if there is an option to "manually" install Filebeat as a service.

Could you also test the deb/systemd flow to make sure we fixed it in all cases?

mauri870:
> Could you also test the deb/systemd flow to make sure we fixed it in all cases?

Sure, I was able to build a deb package with `PACKAGES=deb PLATFORMS=linux/amd64 mage package`, I'll spin up a vm to test this case.

mauri870:
@belimawr From my testing with systemd, filebeat creates /var/log/filebeat/*.ndjson files and `journalctl -u filebeat.service` does contain any logs. The proper behavior would be for the logs to go to stderr in the systemd case as well and appear in the journalctl output, right?

belimawr:
> @belimawr From my testing with systemd, filebeat creates /var/log/filebeat/*.ndjson files and `journalctl -u filebeat.service` does not contain any logs. The proper behavior would be for the logs to go to stderr in the systemd case as well and appear in the journalctl output, right?

Yes, I believe this is the correct behaviour: when running under systemd, the logs go to stderr and can be seen using journalctl.

The best approach to be sure is to test an older version :sweat_smile: 

I also saw your last comment, the behaviour looks correct.

logp/config.go
```diff
@@ -71,15 +71,19 @@ const (
 // Beat is supposed to be run within.
 func DefaultConfig(environment Environment) Config {
 	toFiles := true
+	toStderr := false
 
-	// If running in a container environment, don't write to files by default.
-	if environment == ContainerEnvironment {
+	// For container and systemd environments, we don't write to files by default.
+	switch environment {
+	case ContainerEnvironment, SystemdEnvironment:
```
belimawr, line 78: The fact that we have this logic in two different places, bugs me... Because the FF for 8.15.3 was Tue, I'd get this PR in as it is and backport it if we still have time to make into the release. Then create an issue to re-visit it at some point in the future.

I believe that's the correct function to define the default behaviour, [`createLogOutput`](https://github.com/elastic/elastic-agent-libs/blob/196dace11604a4403f4466812a150ff993c0261c/logp/core.go#L250-L272) should not need the environment switch.

belimawr (approved):

mauri870:
> I also saw your last comment, the behaviour looks correct.

Thanks, I did test with 8.14.3 and the behavior seems consistent with with filebeat main + this fix.

PR #240, opened 2024-10-16
certutil: add functional options

AndersonQ:
## What does this PR do?

Add functional options to New*RootCA and Generate*ChildCert to allow setting a prefix for the CN and multiple DNS names.

Also, it fixes `GenerateChildCert` generating a RSA instead of EC certificate

## Why is it important?

 - set CN prefix: when generating and using multiple certificates, such as to configure mTLS, it's hard to debug any issue if all certificates have the same CN
 - add multiple DNS names: when trying to simulate a real situation, it might require a certificate to have multiple DNSs

## Checklist

## Author's Checklist

- Relates https://github.com/elastic/elastic-agent/issues/4903

### DIFF
--- a/testing/certutil/certutil.go
+++ b/testing/certutil/certutil.go
@@ -40,19 +40,40 @@ type Pair struct {
 	Key  []byte
 }
 
+type configs struct {
+	cnPrefix string
+	dnsNames []string
+}
+
+type Option func(opt *configs)
+
+// WithCNPrefix adds cnPrefix as prefix for the CN.
+func WithCNPrefix(cnPrefix string) Option {
+	return func(opt *configs) {
+		opt.cnPrefix = cnPrefix
+	}
+}
+
+// WithDNSNames adds dnsNames to the DNSNames.
+func WithDNSNames(dnsNames ...string) Option {
+	return func(opt *configs) {
+		opt.dnsNames = dnsNames
+	}
+}
+
 // NewRootCA generates a new x509 Certificate using ECDSA P-384 and returns:
 // - the private key
 // - the certificate
 // - the certificate and its key in PEM format as a byte slice.
 //
 // If any error occurs during the generation process, a non-nil error is returned.
-func NewRootCA() (crypto.PrivateKey, *x509.Certificate, Pair, error) {
+func NewRootCA(opts ...Option) (crypto.PrivateKey, *x509.Certificate, Pair, error) {
 	rootKey, err := ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
 	if err != nil {
 		return nil, nil, Pair{}, fmt.Errorf("could not create private key: %w", err)
 	}
 
-	cert, pair, err := newRootCert(rootKey, &rootKey.PublicKey)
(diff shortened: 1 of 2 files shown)


testing/certutil/certutil.go
```diff
@@ -77,7 +98,36 @@ func NewRSARootCA() (crypto.PrivateKey, *x509.Certificate, Pair, error) {
 // - a Pair with the certificate and its key im PEM format
 //
 // If any error occurs during the generation process, a non-nil error is returned.
-func GenerateChildCert(name string, ips []net.IP, caPrivKey crypto.PrivateKey, caCert *x509.Certificate) (*tls.Certificate, Pair, error) {
+func GenerateChildCert(name string, ips []net.IP, caPrivKey crypto.PrivateKey, caCert *x509.Certificate, opts ...Option) (*tls.Certificate, Pair, error) {
+	priv, err := ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
+	if err != nil {
+		return nil, Pair{}, fmt.Errorf("could not create RSA private key: %w", err)
```
belimawr, line 104: You're calling `ecdsa.GenerateKey`, it looks like a typo in the error message:
```suggestion
		return nil, Pair{}, fmt.Errorf("could not create ECDSA private key: %w", err)
```

testing/certutil/certutil.go
```diff
@@ -77,7 +98,36 @@ func NewRSARootCA() (crypto.PrivateKey, *x509.Certificate, Pair, error) {
...
+			name,
+			ips,
+			priv,
+			&priv.PublicKey,
+			caPrivKey,
+			caCert,
+			opts...)
+	if err != nil {
+		return nil, Pair{}, fmt.Errorf(
+			"could not generate child TLS certificate CA: %w", err)
```
belimawr, line 118: Why is the 'CA' at the end of this error message? I believe "could not generate child TLS certificate: %w" would be enough.
```suggestion
			"could not generate child TLS certificate: %w", err)
```

testing/certutil/certutil.go
```diff
@@ -77,7 +98,36 @@ func NewRSARootCA() (crypto.PrivateKey, *x509.Certificate, Pair, error) {
 // - a Pair with the certificate and its key im PEM format
 //
 // If any error occurs during the generation process, a non-nil error is returned.
-func GenerateChildCert(name string, ips []net.IP, caPrivKey crypto.PrivateKey, caCert *x509.Certificate) (*tls.Certificate, Pair, error) {
+func GenerateChildCert(name string, ips []net.IP, caPrivKey crypto.PrivateKey, caCert *x509.Certificate, opts ...Option) (*tls.Certificate, Pair, error) {
```
belimawr, line 101: Suggestion: Add the algorithm used for the key, like you did with `GenerateRSAChildCert`
```suggestion
func GenerateECDSAChildCert(name string, ips []net.IP, caPrivKey crypto.PrivateKey, caCert *x509.Certificate, opts ...Option) (*tls.Certificate, Pair, error) {
```

AndersonQ: this is kept as is for backwards compatibility ;)

belimawr (approved):

PR #241, opened 2024-10-23
[BUG] do not output error message for connections closed due to timeout

khushijain21:
## What does this PR do?
This PR fixes https://github.com/elastic/beats/issues/40824

## Why is it important?
Avoid polluting log with `use of closed connection` errors due to `idleConnection` timeout

## Checklist


transport/logging.go
```diff
@@ -48,7 +48,12 @@ func LoggingDialer(d Dialer, logger *logp.Logger) Dialer {
 func (l *loggingConn) Read(b []byte) (int, error) {
 	n, err := l.Conn.Read(b)
 	if err != nil && !errors.Is(err, io.EOF) {
-		l.logger.Debugf("Error reading from connection: %v", err)
+		// Check for a closed network connection error
+		if errors.Is(err, net.ErrClosed) {
+			l.logger.Debugf("Connection is closed: %v", err)
```
VihasMakwana, line 53: I think the attached issue mentions that we shouldn't log the error in cases of ErrClosed. But, you're still logging it, with a different message.

VihasMakwana (changes requested):

VihasMakwana (approved):

transport/logging.go
rdner: I think the issue description was a bit misleading, please read my comment here https://github.com/elastic/beats/issues/40824#issuecomment-2431409665

rdner (changes requested):
Blocking this PR until we find a better solution. See https://github.com/elastic/beats/issues/40824#issuecomment-2431409665

rdner (approved):
After some investigation with @khushijain21 we didn't find a better solution other than skipping the log entries for the closed connection error.

This error is still returned and should be handled and logged by the consumer of `loggingConn` type.

PR #249, opened 2024-11-02
Do not throw error if `alterFunc` does not change the original key

khushijain21:
## What does this PR do?
We do not throw error if `alterFunc` does not change the original key. This also improves error message for when key is not found 

## Checklist

### DIFF
--- a/mapstr/mapstr.go
+++ b/mapstr/mapstr.go
@@ -218,12 +218,16 @@ func (m M) AlterPath(path string, mode TraversalMode, alterFunc AlterFunc) (err 
 		if newKey == "" {
 			return fmt.Errorf("replacement key for %q cannot be empty", key)
 		}
-		_, exists := level[newKey]
-		if exists {
-			return fmt.Errorf("replacement key %q already exists: %w", newKey, ErrKeyCollision)
+
+		// if altered key is equal to the original key, skip below delete/put func
+		if newKey != key {
+			_, exists := level[newKey]
+			if exists {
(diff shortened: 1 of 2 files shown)


mapstr/mapstr_test.go
```diff
@@ -1264,6 +1264,22 @@ func TestAlterPath(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:      "alters keys on second level with case-insensitive matching",
```
rdner, line 1268: I think this what this case is testing, right?

```suggestion
			name:      "does not return an error if alterFunc returns the key unchanged",
```

mapstr/mapstr.go
```diff
@@ -218,12 +218,16 @@ func (m M) AlterPath(path string, mode TraversalMode, alterFunc AlterFunc) (err
 		if newKey == "" {
 			return fmt.Errorf("replacement key for %q cannot be empty", key)
 		}
-		_, exists := level[newKey]
-		if exists {
-			return fmt.Errorf("replacement key %q already exists: %w", newKey, ErrKeyCollision)
+
+		// if altered key is equal to the original key, skip below delete/put func
+		if newKey != key {
```
rdner, line 223: I think it would be more readable if it was just:

```go
if newKey == key {
  return nil
}
```

rdner (commented):
Good catch, thanks!
A couple of minor comments.

rdner (approved):
LGTM, just an enhancement comment.

PR #254, opened 2024-11-23
[windows] - Introduce PdhGetRawCounterArrayW

VihasMakwana:
While performing benchmarking, I noticed that by using https://learn.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhgetrawcounterarrayw we can achieve faster results with reduced memory usage. 

This PR adds the methods for PdhGetRawCounterArrayW function.

It also does following thing to improve memory benchmarks:
- Performs sort on the returned array based on instance name. This is helpful as each index of the returned array will correspond to one particular instance (i.e. CPU Core). This helps us save big on memory as we wouldn't need to build a map or perform any segregation based on CPU Cores.
     - The term "instance name" refers to a specific CPU core.
     - For the Processor Information performance counter, there will be `n+1` instance names corresponding to `n` CPU cores.
          - `Processor Information(0)/ % Idle Time` refers to the idle time spent by CPU core `0`
          - `Processor Information(1)/ % Idle Time` refers to the idle time spent by CPU core `1`
          - ...
          - `Processor Information(n)/ % Idle Time` refers to the idle time spent by CPU core `n`
          - `Processor Information(_Total)/ % Idle Time` refers to the average idle time spent for all CPUs.

### DIFF
--- a/helpers/windows/pdh/pdh_query_windows.go
+++ b/helpers/windows/pdh/pdh_query_windows.go
@@ -219,7 +219,7 @@ func (q *Query) GetRawCounterValue(counterName string) (PdhRawCounter, error) {
 	return c, nil
 }
 
-func (q *Query) GetRawCounterArray(counterName string, filterTotal bool) (RawCouterArray, error) {
+func (q *Query) GetRawCounterArray(counterName string, filterTotal bool) (RawCounterArray, error) {
 	if _, ok := q.Counters[counterName]; !ok {
 		return nil, fmt.Errorf("%s doesn't exist in the map; call AddCounter()", counterName)
 	}
--- a/helpers/windows/pdh/pdh_query_windows.go
+++ b/helpers/windows/pdh/pdh_query_windows.go
@@ -80,7 +80,7 @@ func (q *Query) AddEnglishCounter(counterPath string) (PdhCounterHandle, error) 
 }
 
 // AddCounter adds the specified counter to the query.
-func (q *Query) AddCounter(counterPath string, instance string, format string, wildcard bool) error {
+func (q *Query) AddCounter(counterPath string, instance string, format string, wildcard bool, english bool) error {
 	if _, found := q.Counters[counterPath]; found {
 		return nil
 	}
@@ -95,7 +95,12 @@ func (q *Query) AddCounter(counterPath string, instance string, format string, w
 	} else {
 		instanceName = instance
 	}
-	h, err := PdhAddCounter(q.Handle, counterPath, 0)
+	var h PdhCounterHandle
+	if english {
+		h, err = PdhAddEnglishCounter(q.Handle, counterPath, 0)
+	} else {
+		h, err = PdhAddCounter(q.Handle, counterPath, 0)
+	}
 	if err != nil {
 		return err
 	}
--- a/helpers/windows/pdh/pdh_query_windows.go
+++ b/helpers/windows/pdh/pdh_query_windows.go
@@ -203,11 +203,22 @@ func (q *Query) GetCountersAndInstances(objectName string) ([]string, []string, 
 	return UTF16ToStringArray(counters), UTF16ToStringArray(instances), nil
 }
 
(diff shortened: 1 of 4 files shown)


helpers/windows/pdh/pdh_windows.go
```diff
@@ -207,13 +234,43 @@ func PdhGetFormattedCounterValueLong(counter PdhCounterHandle) (uint32, *PdhCoun
 }
 
 // PdhGetRawCounterValue returns the raw value of a given counter.
-func PdhGetRawCounterValue(counter PdhCounterHandle) (*PdhRawCounter, error) {
+func PdhGetRawCounterValue(counter PdhCounterHandle) (PdhRawCounter, error) {
```
leehinman, line 237: Do we need to change the signature?

helpers/windows/pdh/pdh_windows.go
```diff
@@ -207,13 +234,43 @@ func PdhGetFormattedCounterValueLong(counter PdhCounterHandle) (uint32, *PdhCoun
...
+		}
+		buf := make([]byte, bufferSize)
+		if err := _PdhGetRawCounterArray(counter, &bufferSize, &itemCount, &buf[0]); err != nil {
+			return nil, PdhErrno(err.(syscall.Errno))
+		}
+		items := unsafe.Slice((*pdhRawCounterItem)(unsafe.Pointer(&buf[0])), itemCount)
+		ret := make([]PdhRawCounterItem, 0, len(items))
+		for _, item := range items {
+			instance := windows.UTF16PtrToString(item.SzName)
+			if filterTotal && strings.Contains(instance, "_Total") {
```
leehinman, line 260: Do we have to worry if this is a non-English Windows system?  Just wondering if we might have something besides "_Total".

VihasMakwana: For instance name? Nope. The name remains `_Total`. 
I did a quick test on windows VM by changing language to Spanish.

For counter name? yes, the language makes a lot of change.
Hence, I've updated the `AddCounter` function in this PR. PTAL!

helpers/windows/pdh/pdh_query_windows.go
```diff
@@ -95,7 +95,12 @@ func (q *Query) AddCounter(counterPath string, instance string, format string, w
 	} else {
 		instanceName = instance
 	}
-	h, err := PdhAddCounter(q.Handle, counterPath, 0)
+	var h PdhCounterHandle
+	if english {
+		h, err = PdhAddEnglishCounter(q.Handle, counterPath, 0)
+	} else {
+		h, err = PdhAddCounter(q.Handle, counterPath, 0)
+	}
```
VihasMakwana, line 103: These are two different API calls. 
PdhAddCounter is language specific.
PdhAddEnglishCounter is language neutral.

helpers/windows/pdh/pdh_query_windows_test.go
```diff
@@ -193,3 +193,49 @@ func TestUTF16ToStringArray(t *testing.T) {
...
+				assert.NoError(t, err)
+				rawCounters = append(rawCounters, arr)
+			}
+			for i := 0; i < len(rawCounters)-1; i++ {
+				assert.Equalf(t, len(rawCounters[i]), len(rawCounters[i+1]), "returned counters should be equal")
+			}
+			// confirm that each index corresponds to one particular instance (i.e. core)
+			for i := 0; i < len(rawCounters)-1; i++ {
+				for j := 0; j < len(rawCounters[i]); j++ {
+					assert.Equalf(t, rawCounters[i][j].InstanceName, rawCounters[i+1][j].InstanceName, "Instanse name should be equal")
```
mauri870, line 236: Instanse

VihasMakwana: Thanks for catching this.

leehinman (approved):
Thanks for checking on the language pieces.

VihasMakwana:
@mauri870 can you take a look when you have a moment? You left some comments :)

mauri870 (approved):

PR #256, opened 2024-11-29
logp: add rate limited logger

mauri870:
## What does this PR do?

This commit adds support for rate limiting to logp.Logger. This is useful
when you have a lot of log messages and you want to avoid flooding the
destination.

The logger was augmented with the following methods:
    `Throttled` will only log once every period.
    `Sampled` will only log the nth message.
    `Limited` will only log the first n messages.

The implementation is based on x/time/rate.Sometimes.

For all these limiters any message that is not logged will be dropped.

These also synergize with one another, for example:

    log.Sampled(10).
        Throttled(time.Second).
        Info("")`

will log one message every second, and only one out of ten messages.

Likewise,

    log.Limited(10).
        Throttled(time.Second).
        Info("")

will log the first ten messages, and then one message every second.

## Checklist

## Author's Checklist

- [ ]


logp/logger.go
```diff
@@ -81,62 +84,70 @@ func NewInMemory(selector string, encCfg zapcore.EncoderConfig) (*Logger, *bytes
...
 }
 
+func (l *Logger) do(f func()) {
+	if l.limiter != nil {
+		l.limiter.Do(f)
+		return
+	}
+	f()
+}
+
```
mauri870, line 94: Looks like this wrapper is messing with the depth of the stacktraces, I have to address this before merging.

mauri870:
I really like the idea of a rate limited logger, but I think the cost is too high, at least using the approach implemented here.

For a single threaded execution the cost is tolerable, but for mutli-threaded scenarios the contention is too high for a hot path:

```bash
go test -bench='Benchmark.*Logger' -benchmem -run=^$ ./logp | column -t
goos: linux
goarch: amd64
pkg: github.com/elastic/elastic-agent-libs/logp
cpu: Intel(R) Xeon(R) CPU E5-2697A v4 @ 2.60GHz
BenchmarkLogger/default-32                                                                        100000000                                   10.33      ns/op  0         B/op  0  allocs/op
BenchmarkLogger/sampled-32                                                                        17537918                                    69.40      ns/op  0         B/op  0  allocs/op
BenchmarkLogger/throttled-32                                                                      12967281                                    87.35      ns/op  0         B/op  0  allocs/op
BenchmarkLogger/limited-32                                                                        18614289                                    63.58      ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerSynthetic/default-32                                                     1000000000                                  0.0001241  ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerSynthetic/sampled-32                                                     277615                                      13437      ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerSynthetic/throttled-32                                                   236827                                      18124      ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerSynthetic/limited-32                                                     329881                                      9944       ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerRealistic/default-32                                                     1000000000                                  0.0000182  ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerRealistic/sampled-every-4-32                                             347101                                      11292      ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerRealistic/throttled-1-per-second-32                                      544308                                      12081      ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerRealistic/limited-first-10-32                                            1051113                                     5290       ns/op  0         B/op  0  allocs/op
BenchmarkConcurrentLoggerRealistic/sampled-1-out-of-2-limited-first-10-throttled-1-per-second-32  290791                                      20145      ns/op  0         B/op  0  allocs/op
PASS                                                                                                                                                           
ok  
```

PR #257, opened 2024-12-02
chore: replace deprecated Syscall* with SyscallN

mauri870:
## What does this PR do?

Replace deprecated `syscall.Syscall*` functions with `syscall.SyscallN`, a variadic version that can handle any number of arguments.

## Why is it important?

Removes a deprecated function, has no side effects since the numbered variants are based on [SyscallN](https://github.com/golang/go/blob/c5adb8216968be46bd11f7b7360a7c8bde1258d9/src/runtime/syscall_windows.go#L454-L488).

## Checklist

### DIFF
--- a/helpers/windows/pdh/zpdh_windows.go
+++ b/helpers/windows/pdh/zpdh_windows.go
@@ -72,7 +72,7 @@ var (
 )
 
 func _PdhOpenQuery(dataSource *uint16, userData uintptr, query *PdhQueryHandle) (errcode error) {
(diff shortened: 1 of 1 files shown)


leehinman:
This is a generated file.  Can we fix the generator?

mauri870:
> This is a generated file. Can we fix the generator?

Oh, I didn't notice this was autogenerated, thanks. I'll look into it.

mauri870:
Aha, I knew I had seen this before. It turns out I modified mkwinsyscall in Go some time ago to use SyscallN only if the number of arguments exceeds 16. I believe it was decided this way to avoid unnecessary changes on users of this script for code generation.

https://go-review.googlesource.com/c/sys/+/518995

Closing this as not needed. Thanks.

PR #259, opened 2024-12-03
Initialize the interval trigger

nicholasberlin:
## What does this PR do?

Without lastRotate initialized, an interval trigger will always trigger a rotation the first time its callback is executed. To avoid the extra rotation this PR initializes `lastRotate` on creation.

## Why is it important?

Users are having a bad time when setting `rotateonstartup: false` and an interval timer in the logging configuration. They correct expect that a new log will *not* be created on startup but the uninitialized interval timer would create one anyway.

## Checklist

- [ ] ~I have commented my code, particularly in hard-to-understand areas~

### DIFF
--- a/file/helper_other.go
+++ b/file/helper_other.go
@@ -42,6 +42,7 @@ func SyncParent(path string) error {
 	parent := filepath.Dir(path)
 	f, err := os.Open(parent)
 
+	//lint:ignore nilerr failing open is ok
 	if err != nil {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


fearful-symmetry:
@nicholasberlin do we have a test for this behavior? If not, we should probably add one.

nicholasberlin:
> @nicholasberlin do we have a test for this behavior? If not, we should probably add one.

Done: [8686395](https://github.com/elastic/elastic-agent-libs/pull/259/commits/8686395247f4342a4e308caacf29424629eb2243)

I've also verified this by hand with packetbeat.

nicholasberlin:
@fearful-symmetry ready for review

fearful-symmetry (approved):

PR #264, opened 2024-12-31
Update TLS min version to 1.1, set default min to 1.2

michel-laterman:
## What does this PR do?

Set the min TLS version to 1.1, set the default min to 1.2

## Why is it important?

Support TLSv1.0 and TLSv1.1 is being removed/restricted in the upcoming 9.0 release.

If using these versions is still required; for example on the 8.x backports the importing program should call `tlscommon.SetInsecureDefaults()` to restore support for handling TLSv1.0/1.1.

## Checklist

- ~~I have commented my code, particularly in hard-to-understand areas~~
- ~~I have added tests that prove my fix is effective or that my feature works~~

### DIFF
--- a/transport/tlscommon/versions_default.go
+++ b/transport/tlscommon/versions_default.go
@@ -31,14 +31,14 @@ const (
 
 var (
 	// TLSVersionMin is the min TLS version supported.
-	TLSVersionMin = TLSVersion10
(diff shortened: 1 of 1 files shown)


michel-laterman:
This is a duplicate of https://github.com/elastic/elastic-agent-libs/pull/121 😬

PR #279, opened 2025-02-12
feat(fips): Use nop keystore implementation

michel-laterman:
## What does this PR do?

Keystore.Factory and other constuctor functions will return a NopKeystore if the requirefips build tag is set. The NopKeystore will will return ErrKeyDoesntExists for any retrievals, and ErrKeystoreUnsupported for create/save/store operations. The NopKeystore uses the same existing constructor name as the regular file keystore so it can be used in-place by callers.

Minor refactoring to move unchanged constructors, and the `Factory` method to `keystore.go`, renamed `file_keystore.go` to `file_keystore_nofips.go`  to give an addition indicator to developers.

## Why is it important?

Current keystore implementation is not FIPS compliant.

## Checklist

- ~~I have added tests that prove my fix is effective or that my feature works~~


keystore/keystore.go
```diff
@@ -37,8 +39,24 @@ var (
 
 	// ErrNotWritable is returned when the keystore is not writable
 	ErrNotListing = errors.New("the configured keystore is not listing")
+
+	// ErrKeystoreDisabled is returned on create, save, and store operations if built with the requirefips flag
+	ErrKeystoreDisabled = fmt.Errorf("keystore disabled in FIPS mode: %w", errors.ErrUnsupported)
```
michel-laterman, line 44: We could also use `ErrNotWritable` in place of this error

keystore/keystore.go
```diff
@@ -118,3 +142,56 @@ func AsListingKeystore(store Keystore) (ListingKeystore, error) {
...
+// NewFileKeystoreWithPassword return a new File based keystore or an error, allow to define what
+// password to use to create the keystore.
+func NewFileKeystoreWithPassword(keystoreFile string, password *SecureString) (Keystore, error) {
+	return NewFileKeystoreWithPasswordAndStrictPerms(keystoreFile, password, false)
+}
+
+// randomBytes return a slice of random bytes of the defined length
+func randomBytes(length int) ([]byte, error) {
+	r := make([]byte, length)
+	_, err := rand.Read(r)
```
mauri870, line 190: can you do  `io.ReadFull(rand, r)` here? Read is not guaranteed to return exactly `length`.

keystore/file_keystore_fips.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/elastic/elastic-agent-libs/config"
+)
+
+// NopKeystore is used in FIPS mode to a beats keystore.
+type NopKeystore struct{}
+
+// NewFileKeystoreWithPasswordAndStrictPerms returns a NopKeystore.
+//
+// It is using an exising, exported function call name so callers do not need to change the function call, just the build tag.
+func NewFileKeystoreWithPasswordAndStrictPerms(keystoreFile string, password *SecureString, strictPerms bool) (Keystore, error) {
```
mauri870, line 32: Should we have tests for the fips version as well? If the  code here has a bug but we don't test with requirefips we'll only catch it when building beats or elastic-agent with fips enabled.

mauri870 (approved):
Tentatively approved. Please wait for at least one more review before merging.

keystore/keystore.go
pkoutsovasilis: @mauri870 I do like the idea, do you meant to write `io.ReadFull(rand.Reader, randBytes)`? I actually like this proposal 🙂

keystore/file_keystore_fips.go
michel-laterman: Added tests

michel-laterman:
Tests will only succeed once https://github.com/elastic/elastic-agent-libs/pull/281 is merged

pkoutsovasilis (commented):
@michel-laterman please tell me if I am missing something, but as far as I can tell, the current keystore implementation should be FIPS compliant:
* pbkdf2 with SHA512 is approved
* AES with GCM is approved

I can see all the crypto related calls that we use in the code [here](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/UserGuide.md).

With all that said, is the keystore implementation not FIPS compliant? 🤔

pkoutsovasilis (commented):
@michel-laterman can [this unit-test failure](https://buildkite.com/elastic/elastic-agent-libs/builds/599#01951f71-b3b0-419e-af92-ac7a0c242e06/125-1438) happen because FIPS enabled code is not decrypting correctly the blocks of both PKCS1 and PKCS8 keys? 🤔

michel-laterman:
The TLS tests have been restricted for now, #281 fixes the issues with tests.
Once I have a second approval, and #281 has been merged I'll merge this PR

pkoutsovasilis (approved):
I would love this change to be as simple as returning always an error from NewFileKeystoreWithPasswordAndStrictPerms when it is compiled for FIPS. However I understand that due to external dependencies that do not handle appropriately an error there an alternative approach, such as the one of this PR is needed. Thus LGTM

simitt (commented):
@michel-laterman instead of implementing a noop keystore, have you thought of removing the command from the fips compliant binaries? 
Something similar to what was done for the APM Server https://github.com/elastic/apm-server/pull/15545?

michel-laterman:
I have an alternate approach (as @kruskall suggested) that needs reviews: https://github.com/elastic/beats/pull/42846

PR #281, opened 2025-02-18
Add FIPS TLS encrypted private key tests

### LINKED ISSUE #280: [test][FIPS] Add unit test for encrypted private keys being unsupported
Follow up from https://github.com/elastic/elastic-agent-libs/pull/278 to add a test that support for encrypted private keys are not supported in FIPS mode.

michel-laterman:
## What does this PR do?

Add a test to ensure that when if FIPS mode attempting to decrypt an encrypted private key will result in errors.ErrUnsuported. Change the ReadPEM method to return a joined error so that an encrypted block will return an error instead of just having the "no PEM blocks" error.

## Why is it important?

FIPS functionality needs tests.

## Checklist

### Example test run

```
tlscommon|fips-test-private-key ⇒ go test -tags=requirefips -v -race ./...
=== RUN   TestFIPSCertificateAndKeys
=== RUN   TestFIPSCertificateAndKeys/embed
=== RUN   TestFIPSCertificateAndKeys/embed_small_key
=== RUN   TestFIPSCertificateAndKeys/embed_encrypted_PKCS#1_key
=== RUN   TestFIPSCertificateAndKeys/embed_PKCS#8_key
=== RUN   TestFIPSCertificateAndKeys/From_disk
--- PASS: TestFIPSCertificateAndKeys (0.01s)
    --- PASS: TestFIPSCertificateAndKeys/embed (0.00s)
    --- PASS: TestFIPSCertificateAndKeys/embed_small_key (0.00s)
    --- PASS: TestFIPSCertificateAndKeys/embed_encrypted_PKCS#1_key (0.00s)
    --- PASS: TestFIPSCertificateAndKeys/embed_PKCS#8_key (0.00s)
    --- PASS: TestFIPSCertificateAndKeys/From_disk (0.00s)
PASS
ok  	github.com/elastic/elastic-agent-libs/transport/tlscommon	1.274s
```

## Related issues

- Closes #280


transport/tlscommon/ca_pinning_test.go
```diff
@@ -15,6 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
+//go:build !requirefips
+
```
belimawr, line 19: The only thing I didn't understand is why most of the tests won't run when FIPS is enabled.
Do they break with FIPS enabled?

michel-laterman: Some of them break, but this is mostly so we can limit the `requirefips` tests to only test the different behaviour

simitt (approved):
@michel-laterman and I discussed to replace hardcoded test certs with generated ones as a follow up https://github.com/elastic/elastic-agent-libs/issues/282 (out of scope for currently planned work).

transport/tlscommon/ca_pinning_test.go
simitt: @michel-laterman I don't think that we should generally exclude non-system tests from running when in fips mode. What we had discussed was to exclude the expensive system tests that might test certain specifics that aren't focused on FIPS relevant parts; instead we want to have some focused system tests for FIPS covering TLS, crypto, pgp,.. 

Would be worth looking into why these tests are failing rather than disabling them.

belimawr: I agree with @simitt, because FIPS changes some cryptographic libraries it is better to have the full suit of tests dealing with crypto running and only add build tags to the ones with different behaviour.

Also, IIRC the `elastic-agent-libs` CI is rather fast, so I don't see any issues with running the tests twice, one for each build.

simitt (dismissed):
@michel-laterman I created a PR against your branch to get rid of the hardcoded certificates https://github.com/michel-laterman/elastic-agent-libs/pull/1. Please review and let me know if you are ok with merging this into your branch before moving on with this PR.

transport/tlscommon/tls_nofips_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+// TestNoFIPSCertificateAndKeys tests that encrypted private keys are supported in none FIPS mode
+func TestNoFIPSCertificateAndKeys(t *testing.T) {
+	t.Run("embed encrypted PKCS#1 key", func(t *testing.T) {
+		// Create a dummy configuration and append the CA after.
+		password := "abcd1234"
+		key, cert := makeKeyCertPair(t, blockTypePKCS1Encrypted, password)
+		cfg, err := load(`enabled: true`)
+		require.NoError(t, err)
+		cfg.Certificate.Certificate = string(cert)
```
simitt: The string conversion is not necessary here - leftover from my changes

transport/tlscommon/tls_nofips_test.go
```diff
@@ -0,0 +1,117 @@
...
+		cfg.Certificate.Key = string(key)
```
simitt: same as above

transport/tlscommon/tls_nofips_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	t.Run("embed PKCS#8 key", func(t *testing.T) {
+		// Create a dummy configuration and append the CA after.
+		password := "abcd1234"
+		key, cert := makeKeyCertPair(t, blockTypePKCS8Encrypted, password)
+		cfg, err := load(`enabled: true`)
+		require.NoError(t, err)
+		cfg.Certificate.Certificate = string(cert)
+		cfg.Certificate.Key = string(key)
+		cfg.Certificate.Passphrase = password
```
simitt, line 55: same regarding the string conversion

transport/tlscommon/tlscommon_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/elastic-agent-libs/config"
+	"github.com/elastic/go-ucfg"
+	"github.com/elastic/go-ucfg/json"
+	"github.com/elastic/pkcs8"
+)
+
+const (
+	testCert = `-----BEGIN CERTIFICATE-----
```
simitt: Is this still necessary? Should also be replaced

simitt (commented):
Left a few nits

transport/tlscommon/tlscommon_test.go
```diff
@@ -104,3 +111,56 @@ func writeTestFile(t *testing.T, content string) string {
...
+		block = &pem.Block{
+			Type:  "PRIVATE KEY",
+			Bytes: b,
+		}
+	case blockTypePKCS1Encrypted:
+		var err error
+		block, err = x509.EncryptPEMBlock(rand.Reader, "RSA PRIVATE KEY", x509.MarshalPKCS1PrivateKey(key), []byte(password), x509.PEMCipherAES256) //nolint:staticcheck // we need to support encrypted private keys
+		require.NoError(t, err)
+	case blockTypePKCS8Encrypted:
+		//TODO: this uses an elastic implementation of pkcs8 as the stdlib does not support password protected pkcs8
```
belimawr, line 145: Why a `TODO` comment? Shouldn't it be just the comment explaining why we use our custom implementation?
```suggestion
		// This uses an elastic implementation of pkcs8 as the stdlib does not support password protected pkcs8
```

belimawr (approved):
The code looks good, but there is a `TODO` that was left behind. It would be better to remove it before merging.

PR #283, opened 2025-02-21
fix(mapstr): cast inner maps to M on Clone

mauri870:
## What does this PR do?

PR https://github.com/elastic/elastic-agent-libs/pull/262 changed the Clone behavior to preserve the type of inner `map[string]interface{}` instead of casting them to M. This PR reverts this behavior since we rely on inner maps being casted to M in beats. This was not tested before, that is why we only caught it now. Additionally, add tests to ensure this behavior is preserved.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent-libs/pull/262

### DIFF
--- a/mapstr/mapstr.go
+++ b/mapstr/mapstr.go
@@ -169,7 +169,7 @@ func cloneMap(dst, src M) {
 			dst[k] = d
 			cloneMap(d, v)
 		case map[string]interface{}:
-			d := make(map[string]interface{}, len(v))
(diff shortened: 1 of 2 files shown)


leehinman (approved):

efd6 (approved):
I think this is unfortunate and a result of poorly constructed tests in libbeat/processors/add_cloud_metadata, but approving.

PR #290, opened 2025-03-06
Add support for `with_fields` configuration

khushijain21:
## What does this PR do?
It adds with_fields config to logging section - which allows for additional structured context to the logs emitted by beat process
For https://github.com/elastic/ingest-dev/issues/5139

## Checklist

## Related issues

- Part of https://github.com/elastic/ingest-dev/issues/5139


khushijain21:
The docs are not updated with this config because it is introduced to solve an internal problem and is kept private initially.

logp/core_test.go
```diff
@@ -707,7 +707,13 @@ func TestConfigureWithCore(t *testing.T) {
...
 		zapcore.AddSync(&b),
 		zapcore.InfoLevel)
-	err := ConfigureWithCore(Config{}, core)
+
+	config := Config{
+		WithFields: map[string]any{
+			"component": "elastic-agent",
+		},
+	}
+	err := ConfigureWithCore(config, core)
```
VihasMakwana, line 716: Can you add this is another test case? I want to make sure that the logger works in both cases (with or without `with_fields` and we don't wanna cause a regressions.

VihasMakwana (dismissed):

logp/core.go
```diff
@@ -164,6 +164,15 @@ func ConfigureWithCore(loggerCfg Config, core zapcore.Core) error {
 	)
 
 	level = zap.NewAtomicLevelAt(loggerCfg.Level.ZapLevel())
+
+	if loggerCfg.WithFields != nil {
+		var fields = []zapcore.Field{}
+		for key, value := range loggerCfg.WithFields {
```
mauri870, line 170: If we know the length of the slice beforehand we can preallocate.

```go
fields := make([]zapcore.Field, len(loggerCfg.WithFields)
```

mauri870 (dismissed):

mauri870 (approved):

PR #296, opened 2025-03-17
config: protect global configOpts with a mutex

mauri870:
There is a race condition in configOpts when one goroutine calls OverwriteConfigOpts while another reads the variable. This is common, as almost every method in config.C reads this global variable.

Fix the data race by using a mutex. Since this variable is mostly read, optimize it with a sync.RWMutex.

## Checklist

## Related issues

- Relates to https://github.com/elastic/beats/pull/43294.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


leehinman (approved):
Thank You

VihasMakwana (approved):

PR #297, opened 2025-03-19
[Chore] Add functions required for single logger instance

khushijain21:
## What does this PR do?
Required for https://github.com/elastic/ingest-dev/issues/5251

This PR adds some existing methods (on `logp` package) but with `Local` appended to the method name. These  are required by  `libbeat` to use single logger instance. It does not remove any existing methods not to break any code dependent on the `logp` library

## Checklist

## Related issues

- Part of https://github.com/elastic/ingest-dev/issues/5251


logp/core.go
```diff
@@ -202,6 +202,40 @@ func ConfigureWithCore(loggerCfg Config, core zapcore.Core) error {
 	return nil
 }
 
+func ConfigureWithCoreLocal(loggerCfg Config, core zapcore.Core) (*Logger, error) {
```
rdner, line 205: Also needs docs.

logp/core.go
```diff
@@ -256,6 +290,36 @@ func ConfigureWithTypedOutput(defaultLoggerCfg, typedLoggerCfg Config, key, valu
 	return nil
 }
 
+func ConfigureWithTypedOutputLocal(defaultLoggerCfg, typedLoggerCfg Config, key, value string, outputs ...zapcore.Core) (*Logger, error) {
```
rdner, line 293: Also needs docs.

logp/configure/logging.go
```diff
@@ -112,6 +112,37 @@ func LoggingWithTypedOutputs(beatName string, cfg, typedCfg *config.C, logKey, k
...
+	}
+
+	// Make sure we're always running on the same log level
+	typedLogpConfig.Level = config.Level
+	typedLogpConfig.Selectors = config.Selectors
+
+	// If the name has not been configured, make it {beatName}-events-data
+	if typedLogpConfig.Files.Name == defaultName {
+		typedLogpConfig.Files.Name = beatName + "-events-data"
+	}
```
rdner, line 142: Looks like the entire code region matches the `LoggingWithTypedOutputs` function. This should be extracted in a separate non-exported function used in both.

khushijain21: If we eventually want to remove `LoggingWithTypedOutputs` and make entire `logp` package return only local logging instances - it would be better to duplicate code than extract the common code region.

logp/core.go
```diff
@@ -256,6 +290,36 @@ func ConfigureWithTypedOutput(defaultLoggerCfg, typedLoggerCfg Config, key, valu
...
+	}
+
+	sink = &typedLoggerCore{
+		defaultCore: sink,
+		typedCore:   typedCore,
+		key:         key,
+		value:       value,
+	}
+
+	sink = selectiveWrapper(sink, selectors)
```
AndersonQ, line 316: Same here, it seems a duplication of most of the code from `ConfigureWithTypedOutput`. Ideally it'd be extracted to a unreported function and both `ConfigureWithTypedOutput` and `ConfigureWithTypedOutputLocal` would call it.

logp/configure/logging.go
rdner: I agree in this case. I would put a comment (TODO or deprecation) on LoggingWithTypedOutputs for visibility.

AndersonQ (commented):
@khushijain21 for the function which need docs, if the doc would be the same as for the non-local version you can refer to the non-local and point out the difference. Like that:

```
// SomeFunction performs a specific operation and logs relevant information using the global logger.
//
// It takes an input string and processes it, logging the input and the result of the processing.
func SomeFunction(input string) string {


// SomeFunctionLocal is the same as SomeFunction but returns a logger instance instead.
func SomeFunctionLocal(input string, logger *log.Logger) (string, *log.Logger) {
```

khushijain21:
With new local methods, we will lose the `observerLogs` parameter available on global config. 

Unless we change [Logger](https://github.com/elastic/elastic-agent-libs/blob/main/logp/logger.go#L35) struct to include 
```
type Logger struct {
	logger       *zap.Logger
	sugar        *zap.SugaredLogger
	observerLogs *observer.ObservedLogs. // additional
}
```

This would change many API's that depend on `Logger`.

Instead, users can get similar functionality with this
```
	observed, zapLogs := observer.New(zapcore.DebugLevel)
	observer, err := logp.ConfigureWithCoreLocal(defaultConfig, observed)
```

Any thoughts? I prefer going with the second option.

khushijain21:
> [observerLogs](https://github.com/elastic/elastic-agent-libs/blob/main/logp/core.go#L63) was just for testing purposes and validating logs after a test? Is it used elsewhere?

Yes it is used only for testing. It is only used in 9 files in beats https://github.com/search?q=repo%3Aelastic%2Fbeats%20ObserverLogs&type=code

> If we change all our interfaces to accept a local logger, testing should also do so and everything should be recorded in the local logger instead.

Yes, so users can create an observer core and access "testing logs" on the local logger. As described on the second part here https://github.com/elastic/elastic-agent-libs/pull/297#issuecomment-2736633556

rdner (approved):

logp/configure/logging.go
```diff
@@ -79,6 +79,7 @@ func LoggingWithOutputs(beatName string, cfg *config.C, outputs ...zapcore.Core)
 	return logp.ConfigureWithOutputs(config, outputs...)
 }
 
+// Deprecated (moving towards localized loggers): Use logp.LoggingWithTypedOutputsLocal.
 // LoggingWithTypedOutputs applies some defaults then calls ConfigureWithTypedOutputs
```
mauri870, line 83: In order for godoc to report the API as deprecated it needs to be written in a specific way.

See https://go.dev/wiki/Deprecated.

```suggestion
// LoggingWithTypedOutputs applies some defaults then calls ConfigureWithTypedOutputs
//
// Deprecated: Moving towards localized loggers. Use logp.LoggingWithTypedOutputsLocal.
```

logp/core.go
```diff
@@ -151,6 +151,7 @@ func ConfigureWithOutputs(defaultLoggerCfg Config, outputs ...zapcore.Core) erro
 	return nil
 }
 
+// Deprecated (moving towards localized loggers): Use logp.ConfigureWithCoreLocal.
```
mauri870, line 154: Same for this block.

```go
// ConfigureWithCore configures the global logger to use the passed in
// core.  It is assumed that an output has already been defined with
// the core and a new one should not be created.  The loggerCfg is
// only used to set selectors and level.  This is useful if a part of
// your code uses logp but the log output is already handled.  Normal
// use cases should use Configure or ConfigureWithOutput.
//
// Deprecated: Prefer localized loggers. Use logp.ConfigureWithCoreLocal.

**Comment by leehinman** [comment]
I'd really like @belimawr to take a look before we merge.  I think it is safe since we are only adding functions, but logp has a history of unexpected behavior.

File: logp/logger.go [hunk 7]
```diff
@@ -53,12 +56,41 @@ func NewLogger(selector string, options ...LogOption) *Logger {
...
+func NewProductionLogger(selector string, options ...LogOption) (*Logger, error) {
+	log, err := zap.NewProduction(options...)
+	log = log.Named(selector)
+	if err != nil {
+		return nil, err
+	}
+	return &Logger{log, log.Sugar()}, nil
+}
+
+// NewProductionLogger returns a development suitable logp.Logger
```
belimawr, line 69:
```suggestion
// NewDevelopmentLogger returns a development suitable logp.Logger
```

logp/logger.go
```diff
@@ -53,12 +56,41 @@ func NewLogger(selector string, options ...LogOption) *Logger {
...
+func NewDevelopmentLogger(selector string, options ...LogOption) (*Logger, error) {
+	log, err := zap.NewDevelopment(options...)
+	log = log.Named(selector)
+	if err != nil {
+		return nil, err
+	}
+	return &Logger{log, log.Sugar()}, nil
+}
+
+// NewProductionLogger returns a development suitable logp.Logger
```
belimawr, line 79:
```suggestion
// NewTestingLogger returns a testing suitable logp.Logger
```

logp/logger.go
```diff
@@ -53,12 +56,41 @@ func NewLogger(selector string, options ...LogOption) *Logger {
...
+func NewTestingLogger(t *testing.T, selector string, options ...LogTestOption) *Logger {
+	log := zaptest.NewLogger(t, options...)
```
belimawr, line 81: Overall I believe it would be helpful to clarify the difference between the 'development' and 'testing' loggers.

I see they use different zap functions, but what does that mean to the caller of `NewTestingLogger` and `NewDevelopmentLogger`?

logp/logger.go
```diff
@@ -81,6 +113,34 @@ func NewInMemory(selector string, encCfg zapcore.EncoderConfig) (*Logger, *bytes
...
+	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	encoder := zapcore.NewConsoleEncoder(encoderConfig)
+
+	core := zapcore.NewCore(
+		encoder,
+		zapcore.Lock(zapcore.AddSync(&buff)),
+		zap.NewAtomicLevelAt(zap.DebugLevel))
+	ecszap.ECSCompatibleEncoderConfig(ConsoleEncoderConfig())
+
+	logger, _ := NewDevelopmentLogger(
```
belimawr, line 135: The original function calls `NewLogger`, why the change? What is the advantage of using `NewDevelopmentLogger` here?

khushijain21: The `NewLogger` depends on the [global logger](https://github.com/elastic/elastic-agent-libs/blob/main/logp/logger.go#L53) initialized with this package - now it returns a local logger instance - which is why this change

logp/logger.go
khushijain21: so `NewTestingLogger` is functionally similar to [TestingSetup](https://github.com/elastic/elastic-agent-libs/blob/main/logp/core.go#L300) method.

and `NewDevelopmentLogger` is similar to [DevelopmentSetup](https://github.com/elastic/elastic-agent-libs/blob/main/logp/core.go#L285)


`NewTestingLogger` integrates well with go tests and only logs message when the test fails or when `go test -v` (verbose) flag is used.

I'll add more comments in the code

belimawr: I see, thanks for clarifying! I don't think I had ever seen the `TestingSetup` function, I've always used `DevelopmentSetup` :sweat_smile:

logp/logger.go
belimawr: I see it now. Thanks for clarifying!

rdner (approved):

PR #302, opened 2025-03-28
Enable GODEBUG=fips140=only tests in buildkite.

michel-laterman:
## What does this PR do?

Enable unit tests with GODEBUG=fips140=only and -tags=requirefips in buildkite. This is a basic check to make sure we don't accidently add non-compliant crypto when in FIPS mode.
When these test are ran, the encrypted private key tests in transport/tlscommon are skipped, the -tags=requirefips tests check that an error is returned, however if fips140=only is set the test will panic.
Add a new V2 file keystore that is used with requirefips. This implementation uses NewGCMWithRandomNonce. File keystore support is currently disabled in FIPS artifacts we produce.

## Why is it important?

We want automated testing to ensure we don't accidentally introduce non-compliant code

## Checklist

- [c] I have added tests that prove my fix is effective or that my feature works


michel-laterman:
Requires update to go 1.24: https://github.com/elastic/elastic-agent-libs/pull/303

.buildkite/pipeline.yml
```diff
@@ -36,6 +37,19 @@ steps:
     artifact_paths:
       - "junit-*.xml"
 
+    # Run unit tests with requirefips tag and GODEBUG=fips140=only
+    # This is a check against accidently adding crypto that breaks FIPS compliance.
```
mauri870, line 41:
```suggestion
    # This is a check against accidentally adding crypto that breaks FIPS compliance.
```

transport/tlscommon/tls_fips_test.go
```diff
@@ -29,6 +31,9 @@ import (
 
 // TestFIPSCertifacteAndKeys tests that encrypted private keys fail in FIPS mode
 func TestFIPSCertificateAndKeys(t *testing.T) {
+	if checkGODEBUG() {
+		t.Skip("GODEBUG=fips140=only detected, avoiding panics")
+	}
```
simitt, line 36: Please add a comment as to why this test needs to be excluded for `fips140=only`. As these are the FIPS tests (not the nonfips ones) it requires a nuanced understanding of the various fips related testing strategies to understand why fips tests are skipped in this scenario.

pkoutsovasilis: +1 to the above, I cannot fully decode why this test under `requirefips` tag requires this check 🙂

transport/tlscommon/tls_fips_test.go
```diff
@@ -46,7 +52,7 @@ func TestFIPSCertificateAndKeys(t *testing.T) {
 
 	t.Run("embed encrypted PKCS#8 key", func(t *testing.T) {
 		// Create a dummy configuration and append the CA after.
-		password := "abcd1234"
+		password := "abcdefg1234567"
```
michel-laterman, line 55: if the password is less then 112 bits we also get a panic

transport/tlscommon/tls_fips_test.go
michel-laterman: added the comment, moved the check to a subtest; and increased password length so the other subtest passes

transport/tlscommon/tls_fips_test.go
```diff
@@ -30,6 +32,10 @@ import (
 // TestFIPSCertifacteAndKeys tests that encrypted private keys fail in FIPS mode
 func TestFIPSCertificateAndKeys(t *testing.T) {
 	t.Run("embed encrypted PKCS#1 key", func(t *testing.T) {
+		if checkGODEBUG() {
+			t.Skip("GODEBUG=fips140=only detected, avoiding panics")
```
pkoutsovasilis, line 36: I still don't quite get why we should skip this. I would expect a `requirefips` tagged build should not get panics. Is there an issue somewhere that we should fix in order for this not to panic and I am not seeing? 🙂

transport/tlscommon/tls_fips_test.go
```diff
@@ -58,3 +64,16 @@ func TestFIPSCertificateAndKeys(t *testing.T) {
...
+	// NOTE: This only checks env var; at the time of writing fips140 can only be set via env
+	// other GODEBUG settings can be set via embedded comments or in go.mod, we may need to account for this in the future.
+	ss := strings.Split(os.Getenv("GODEBUG"), ",")
+	for _, s := range ss {
+		if s == "fips140=only" {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
michel-laterman: `crypto/fips140.Enabled()` also returns true if `fips140=on`

transport/tlscommon/tls_fips_test.go
michel-laterman: `GODEBUG=fips140=only` will panic if any non-FIPS approved algorithm is used.
The purpose of adding the buildkite step with this setting is to add automation so that we do not accidentally add things that break FIPS compliance (unit tests that just test `requirefips` are already running elsewhere)

michel-laterman: To be more clear; in our regular FIPS tests will test that `_, err = LoadTLSConfig(cfg)` returns an error (expected behaviour when we compile for FIPS).

When `GODEBUG=fips140=only` is used the test fails in setup as  `makeKeyCertPair()` panics

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/elastic/elastic-agent-libs
 
-go 1.24.1
+go 1.23
```
pkoutsovasilis, line 3: this shouldn't change right? 🙂

pkoutsovasilis: ok scratch that I saw that you added the version go:build guards

pkoutsovasilis (approved):
LGTM 🙂

PR #306, opened 2025-04-01
feat: downgrade to go 1.23

kruskall:
## What does this PR do?

do not force consumer of the library on go 1.24
revert of https://github.com/elastic/elastic-agent-libs/pull/303 sort of

## Why is it important?

(this is a library so it should support all the supported versions of go)

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -24,8 +24,8 @@ steps:
       - "junit-*.xml"
 
     # Run unit tests with requirefips tag to validate functionality
-  - label: ":linux: Test Linux FIPS"
-    key: test-lin-fips
+  - label: ":linux: Test Linux with requirefips build tag"
+    key: test-lin-fips-tag
     command:
       - ".buildkite/scripts/test.sh"
     env:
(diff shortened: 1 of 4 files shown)


cmacknz:
I would change .go-version as well https://github.com/elastic/elastic-agent-libs/blob/main/.go-version

That's the Go version used in CI, if the go.mod says compatible with go 1.23 that should imply we don't use Go 1.24 features and we are going to need to rely on CI to enforce that (unless this directive does that for us, currently unsure).

cmacknz (approved):

mauri870 (changes requested):
The version 1.23 in the go.mod is not a valid Go release, the first valid go release of 1.23 is 1.23.0. See https://go.dev/issues/62278#issuecomment-1693538776

I did not test this but something like `GOTOOLCHAIN="go1.23+auto" go1.24.1 mod tidy` will probably fail with this commit.

mauri870 (dismissed):

faec:
It looks like this is failing because the FIPS code in the keystore package requires go 1.24: https://github.com/elastic/elastic-agent-libs/blob/091adb0d320ec5776c8850581b3f96230e214f2e/keystore/file_keystore_fips_test.go#L18

Possibly you could fix it by changing the go version for just that test step: https://github.com/elastic/elastic-agent-libs/blob/091adb0d320ec5776c8850581b3f96230e214f2e/.buildkite/pipeline.yml#L34

PR #307, opened 2025-04-01
User Agent: Rearrange the user agent to prioritize displaying the most useful information.

pierrehilbert:
## What does this PR do?

Change the `user_agent` to display information in another order.

From: `Elastic-<binaryNameCapitalized>/<version> (<runtime.GOOS>; <runtime.GOARCH>; <commit>; <buildTime>; <Managed|Standalone>; <Unprivileged>`

To: `Elastic-<binaryNameCapitalized>/<version> (<Managed|Standalone>; <Unprivileged>; <runtime.GOOS>; <runtime.GOARCH>; <commit>; <buildTime>`

## Why is it important?

Some proxy log solutions are stripping the user_agent size and the most useful information are therefore lost.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/useragent/useragent.go
+++ b/useragent/useragent.go
@@ -27,17 +27,13 @@ import (
 func UserAgent(binaryNameCapitalized string, version, commit, buildTime string, additionalComments ...string) string {
 	var builder strings.Builder
 	builder.WriteString("Elastic-" + binaryNameCapitalized + "/" + version + " ")
-	uaValues := []string{
-		runtime.GOOS,
-		runtime.GOARCH,
-		commit,
-		buildTime,
-	}
+	var uaValues []string
 	for _, val := range additionalComments {
 		if val != "" {
 			uaValues = append(uaValues, val)
 		}
 	}
+	uaValues = append(uaValues, runtime.GOOS, runtime.GOARCH, commit, buildTime)
 	builder.WriteByte('(')
 	builder.WriteString(strings.Join(uaValues, "; "))
 	builder.WriteByte(')')
--- a/useragent/useragent_test.go
+++ b/useragent/useragent_test.go
@@ -35,5 +35,5 @@ func TestUserAgent(t *testing.T) {
 	assert.Regexp(t, regexp.MustCompile(`^Elastic-FakeBeat`), ua)
 
 	ua2 := UserAgent("FakeBeat", v, commit, buildTime, "integration_name/1.2.3")
-	assert.Regexp(t, regexp.MustCompile(`; integration_name\/1\.2\.3\)$`), ua2)
+	assert.Regexp(t, regexp.MustCompile(`\(integration_name\/1\.2\.3;`), ua2)
 }


useragent/useragent.go
```diff
@@ -27,17 +27,13 @@ import (
...
-		commit,
-		buildTime,
-	}
+	var uaValues []string
 	for _, val := range additionalComments {
 		if val != "" {
 			uaValues = append(uaValues, val)
 		}
 	}
+	uaValues = append(uaValues, runtime.GOOS, runtime.GOARCH, commit, buildTime)
```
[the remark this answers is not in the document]
pierrehilbert: This is actually the objective here, to rearrange the order.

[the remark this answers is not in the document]
belimawr: Is the user-agent cached some how or do we rebuild it every time we need?

If the goal is to optimise this function/minimise allocations, maybe we can call `builder.WriteString` directly instead.

I'm not against the current implementation, but if the focus is to optimise the function, there might be other points to focus on aside this one append.

belimawr (dismissed):
LGTM, but it's better to wait for @cmacknz's answer.

useragent/useragent.go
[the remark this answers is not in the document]
cmacknz: The actual goal is to remove the build time and remove or shorten the commit. The reason is because this is over the size limit in the proxy logs we use for telemetry queries.

This change implicitly drops those fields by putting them at the end, after the point of truncation. However, this is in no way obvious to people who don't know all this context and it is still a breaking change to any telemetry queries that rely on these fields being present in Beats and other Go binaries that might import this and depend on this format.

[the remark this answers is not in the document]
cmacknz: We are fine with changing the schema of the Beats user agent, but not necessarily in all versions of Beats, and not necessarily in all Go programs that might import this.

So I think regardless of solving this by rearranging things or just outright deleting them, we need some way to opt in to changing the user agent.

We have had a problems in the past where we need to update elastic-agent-libs to pick up a dependency bump but newer releases containing unavoidable breaking changes and I worry this is going to be in this category.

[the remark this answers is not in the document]
cmacknz: (there is also a chance users depend on the format of the user agent in addition to us, so far we have only extended it and not broken the schema of the previously existing fields)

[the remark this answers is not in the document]
nimarezainia: Yes the goal was to drop the commit and the build time. I'm not aware of anyone using these fields. I'm not sure how useful they are from a telemetry standpoint.

VihasMakwana (dismissed):

useragent/useragent.go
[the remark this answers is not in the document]
cmacknz: If we are fine with breaking the current format, then my preference for how to implement this is:

1. We create a second User Agent function. Either we add one with a Short suffix that has the commit+build time deleted completely, or we move the existing one to a Legacy or a Long suffixed version and shrink the default one. This depends on if we want this change to apply to every user of elastic-agent-libs or if we want to only affect Beats. I think having an opt-in Short variant is safer but it is more work to update it everywhere.
2. We write an explicit test that all permutations of the shortened User-Agent fit into the 100 character limit. I suspect we may need to shrink some of the additional agent metadata related to Privileged/Managed/FIPS to do this so lets just deal with that now to make this change one time.

useragent/useragent.go
```diff
@@ -43,3 +43,21 @@ func UserAgent(binaryNameCapitalized string, version, commit, buildTime string,
 	builder.WriteByte(')')
 	return builder.String()
 }
+
+func UserAgentLite(binaryNameCapitalized string, version string, additionalComments ...string) string {
```
cmacknz, line 47:
```suggestion
func UserAgentWithBeatTelemetry(binaryNameCapitalized string, version string, standalone bool, unprivileged bool) string {
```

Let's use a really specific name to communicate the purpose is telemetry, and stop using `additionalComments ...string` as variable arguments to accept the parameters we know the Beat is going to add in the end. Namely this is whether it is managed or unprivileged, you can see these being added here: https://github.com/elastic/beats/blob/7458a19634fc329860963563a23af7780ecfdee9/libbeat/outputs/elasticsearch/elasticsearch.go#L117

```go
	if beat.Manager != nil && beat.Manager.Enabled() {
		info := beat.Manager.AgentInfo()
		if info.ManagedMode == proto.AgentManagedMode_MANAGED {
			comments = append(comments, "Managed")
		} else if info.ManagedMode == proto.AgentManagedMode_STANDALONE {
			comments = append(comments, "Standalone")
		}

		if info.Unprivileged {
			comments = append(comments, "Unprivileged")
		}
	}
```

Arguably if we are going to do this, and intend to only use it in the Beat Elasticsearch output, we should just move this function into Beats.

Looking at other uses of the original UserAgent function in Beats it is used widely in inputs so I think creating a new function here was justified. We don't need or want to change the user agent everywhere unnecessarily, just the one for the ES output. https://github.com/search?q=repo%3Aelastic%2Fbeats%20UserAgent&type=code

useragent/useragent_test.go
```diff
@@ -37,3 +37,11 @@ func TestUserAgent(t *testing.T) {
 	ua2 := UserAgent("FakeBeat", v, commit, buildTime, "integration_name/1.2.3")
 	assert.Regexp(t, regexp.MustCompile(`; integration_name\/1\.2\.3\)$`), ua2)
 }
+
+func TestUserAgentLite(t *testing.T) {
```
cmacknz, line 41: Following the suggestion in https://github.com/elastic/elastic-agent-libs/pull/307/files#r2035942942, add a test to guarantee all permutations of the user agent under the 100 byte limit since we will have limited the set of arguments we can accept and can now test this explicitly.

useragent/useragent.go
cmacknz: FYI We are going to have to additionally extend this to handle adding FIPS to the user agent.

pierrehilbert: I think the objective when @fearful-symmetry added this `additionalComments` variable was to be able to add additional information easily when needed.

If we are changing the behavior (and also using this `WithBeatTelemetry` name), should we also change the existing one by removing the additionalComments?

cmacknz: Adding arbitrary additional strings makes total sense if you do not have a size constraint. We now have a size constraint, and we should make it hard to violate it, otherwise the changes people make with the flexible API today are pointless. This is the case with our own changes for unprivileged and fleet management state.

Removing additionalComments is a breaking API change, so I'd avoid it. We set the user agent for outgoing HTTP requests to systems that are not run on Elastic infrastructure and don't have a size constraint either.

pierrehilbert:
@cmacknz I addressed your comment, could you please review again?

useragent/useragent.go
```diff
@@ -43,3 +43,18 @@ func UserAgent(binaryNameCapitalized string, version, commit, buildTime string,
 	builder.WriteByte(')')
 	return builder.String()
 }
+
+func UserAgentWithBeatTelemetry(binaryNameCapitalized string, version string, mode string, unprivileged string) string {
```
cmacknz, line 47:
```suggestion
func UserAgentWithBeatTelemetry(binaryNameCapitalized string, version string, mode AgentManagementMode, unprivileged bool) string {
```

The mode should probably be an enum so that it isn't totally arbitrary and unprivileged should be a bool. You can pre-emptively add FIPS as a bool in here if you want since we know we are going to need that.

useragent/useragent.go
```diff
@@ -43,3 +43,18 @@ func UserAgent(binaryNameCapitalized string, version, commit, buildTime string,
...
+	uaValues := []string{
+		runtime.GOOS,
+		runtime.GOARCH,
+		mode,
+		unprivileged,
+	}
+	builder.WriteByte('(')
+	builder.WriteString(strings.Join(uaValues, "; "))
+	builder.WriteByte(')')
+	return builder.String()
```
cmacknz, line 59: Since the input is still quite flexible, what I'd suggest is to have this function return an error when the user agent is over 100 bytes. The test doesn't catch all instances of the user agent being too long because that depends on the input.

Then, when you call it you can look at the error and log it but always set the user agent anyway, and the test can just check the function doesn't return an error.

useragent/useragent.go
pierrehilbert: About making unprivileged a bool, I was thinking about it but found it less speaking from a reading perspective in the user-agent itself. 
Should we consider the parameter to be a boolean but have `<privileged | unprivileged>` written in the user-agent when we are building it? Same question goes for fips.

cmacknz: The intent is to standardize the format of the user agent by eliminating the ability of the caller to control the exact text, only the components of the text. There is a size constraint to account for so we should centralize preserving the size constraint into this function and not expose it to callers at all. They shouldn't need to know it exists.

Instead of a bool you could use enums, which would probably make calls to the function easier to read.

pierrehilbert: I added two enums. I let you confirm this is what you were expecting here.

cmacknz (approved):
This looks good, but now our problem is it will quickly be obsoleted when we move to the ES exporter, so we'll have to make an update there. I haven't looked in depth if it's possible to add arbitrary content to the user agent, if it's not, we'll need that first.

graphaelli:
good to merge now?

PR #313, opened 2025-04-28
Add helper to create a sub-registry or return the existing one if present

faec:
When creating a child registry, code often checks if the registry already exists, since creating a duplicate entry will cause a panic. There are two issues with this:

- Most registry creation has extra identical boilerplate to check if a registry exists before creating it
- This boilerplate is not concurrency-safe, because the registry lock is not held continuously; it's possible for two goroutines to both check the same registry name, both observe that it does not exist yet, and both try to create it. This is rare (I don't know of confirmed occurrences) but if it does happen the entire process will panic.

This PR adds a helper that handles both these issues, replacing the usual checks with a one-line call that properly holds the lock between checking for existing registries and creating new ones.

## Checklist

### DIFF
--- a/monitoring/registry.go
+++ b/monitoring/registry.go
@@ -106,29 +106,52 @@ func (r *Registry) Get(name string) Var {
 
 // GetRegistry tries to find a sub-registry by name.
 func (r *Registry) GetRegistry(name string) *Registry {
-	e, err := r.find(name)
-	if err != nil {
-		return nil
+	if v := r.Get(name); v != nil {
+		if reg, ok := v.(*Registry); ok {
+			return reg
+		}
 	}
+	return nil
+}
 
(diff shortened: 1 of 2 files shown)


monitoring/registry.go
```diff
@@ -106,22 +106,52 @@ func (r *Registry) Get(name string) Var {
 
 // GetRegistry tries to find a sub-registry by name.
 func (r *Registry) GetRegistry(name string) *Registry {
-	e, err := r.find(name)
```
faec, line 109: The changes in GetRegistry aren't strictly needed (they're a functional no-op), just some cleanup to remove some duplicate lines while I'm in here

leehinman (approved):

VihasMakwana (approved):

Issue #315, opened 2025-05-02
Gracefully handle os.IsNotExist error when removing UNIX socket file


pkoutsovasilis:
Currently, when attempting to remove an existing UNIX socket file, the code checks for the file’s existence and then calls a remove operation. If the file is deleted between these two operations (for example, due to concurrent cleanup or filesystem behavior), the removal returns an error indicating that the file does not exist.

This results in an unnecessary failure, even though the desired state (the file being absent) is already achieved.

We should adjust the logic to ignore “file does not exist” errors when attempting to remove the UNIX socket, so that this case is handled gracefully rather than causing an error.

PR #316, opened 2025-05-02
fix: ignore os.IsNotExist error when removing stale UNIX socket

pkoutsovasilis:
## What does this PR do?

This PR improves the cleanup logic for removing existing UNIX socket files before binding a new listener.

Previously, if the socket file was deleted between the `os.Stat` and `os.Remove` calls, `os.Remove` would return an `os.IsNotExist` error and the listener setup would fail unnecessarily.

This PR updates the logic to **ignore `os.IsNotExist` errors from `os.Remove`**, preventing spurious failures in this scenario. It also adds a debug log message when `os.Stat` reports an error, to aid troubleshooting.

## Why is it important?

This change improves the robustness of UNIX socket listener creation. Without this change, users may encounter hard-to-debug startup failures if the socket file is unexpectedly absent between `os.Stat` and `os.Remove`.

## Checklist

## Author's Checklist

N/A

## Related issues

- Closes https://github.com/elastic/elastic-agent-libs/issues/315

### DIFF
--- a/api/make_listener_posix.go
+++ b/api/make_listener_posix.go
@@ -26,9 +26,10 @@ import (
 	"os"
 
 	"github.com/elastic/elastic-agent-libs/api/npipe"
+	"github.com/elastic/elastic-agent-libs/logp"
 )
 
-func makeListener(cfg Config) (net.Listener, error) {
+func makeListener(log *logp.Logger, cfg Config) (net.Listener, error) {
 	if len(cfg.User) > 0 {
 		return nil, errors.New("specifying a user is not supported under this platform")
 	}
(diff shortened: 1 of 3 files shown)


api/make_listener_posix.go
```diff
@@ -48,7 +49,8 @@ func makeListener(cfg Config) (net.Listener, error) {
 
 	if network == unixNetwork {
 		if _, err := os.Stat(path); !os.IsNotExist(err) {
```
leehinman, line 51: How about removing the `Stat` call as well?  Then we remove the whole pattern that resulted in the bug.

pkoutsovasilis: since we don't use the stat info then yes that will work, thus I pushed it here 715ce2bb209014fdbf6ee0a31d0b4eddb41bdad6

leehinman (approved):

VihasMakwana (approved):

mauri870 (approved):

PR #323, opened 2025-06-05
fips: Remove specific keystore handling

simitt:
## What does this PR do?
Revert fips specific behavior with `NewGCMWithRandomNonce` as this is not implemented by the go/microsoft toolchain.

## Why is it important?

In the past the keystore functionality was changed to implement the newer crypto function `NewGCMWithRandomNonce` rather than `NewGCM` in fips mode. This is not currently implemented though by microsoft/go (https://github.com/microsoft/go/issues/1648) and therefore cannot be used when built with this toolchain for FIPS. 

The keystore is already disabled in the downstream binaries (beats, fleet-server, elastic-agent, apm-server), so disabling the keystore for fips mode here doesn't lead to any negative effect.  

## Checklist


simitt:
From `go-1.25` onwards, `microsof-go` does support `crypto.NewGCMWithRandomNonce`, see https://github.com/microsoft/go/issues/1648#issuecomment-2988621617. As we usually keep up-to-date with latest go versions and keystore functionality is currently not used in dependencies when in FIPS mode, closing this PR to keep the logic as is.

PR #324, opened 2025-06-12
Add nop logger

belimawr:
## What does this PR do?

This PR adds a no-op logger

## Why is it important?

We're moving away from using global logger instances, this is making more functions to require a logger as an argument, for testing having a no-op logger is very helpful

## Checklist

~~## Author's Checklist~~
~~## Related issues~~

### DIFF
--- a/logp/logger.go
+++ b/logp/logger.go
@@ -53,6 +53,12 @@ func NewLogger(selector string, options ...LogOption) *Logger {
 	return newLogger(loadLogger().rootLogger, selector, options...)
 }
 
+// NewNopLogger returns a no-op logger
+func NewNopLogger() *Logger {
+	logger := zap.NewNop()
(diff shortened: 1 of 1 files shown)


logp/logger.go
AndersonQ, line 1: as it's intended for tests, it'd be better on `logp/logptest/`

logp/logger.go
```diff
@@ -53,6 +53,12 @@ func NewLogger(selector string, options ...LogOption) *Logger {
 	return newLogger(loadLogger().rootLogger, selector, options...)
 }
 
+// NewNopLogger returns a no-op logger
+func NewNopLogger() *Logger {
+	logger := zap.NewNop()
+	return &Logger{logger, logger.Sugar()}
+}
+
```
AndersonQ, line 61: as it's intended for tests, it'd be better on `logp/logptest/`

PR #326, opened 2025-06-25
Add support for `logging.event_data` in beatreceivers

khushijain21:
## What does this PR do?
This PR adds support for `logging.event_data` for beatreceivers running from EA. The idea is that all documents containing sensitive information should be logged to separate `-event.log` file. 

Beats' adds a field called `log.type:event` to log messages containing sensitive information [Ref](https://github.com/khushijain21/beats/blob/gcppubsub/libbeat/outputs/elasticsearch/client.go#L512) and 
[typedLoggerCore](https://github.com/khushijain21/elastic-agent-libs/blob/eventLogging/logp/typedloggercore.go#L105) looks for the presence of this key-value pair. If exists, it logs to `-event.log` file.

Documents rejected by ES-exporter also fall in this category. We add a new field called `message` on `typedLogger`. If a log entry contains this message, it is "also" logged using `typedLogger`. 

## Checklist

## Related issues

- Closes https://github.com/elastic/beats/issues/44603


logp/core.go
```diff
@@ -254,6 +256,51 @@ func ConfigureWithCoreLocal(loggerCfg Config, core zapcore.Core) (*Logger, error
 	return newLogger(root, ""), nil
 }
 
+// ConfigureOTelCore returns beatreceiver suitable logger
+// If a log entry contains typedMsg, it is logged using core generated by typedCfg
+func ConfigureOTelCore(beatName string, logCfg, typedCfg *config.C, typedMsg string, core zapcore.Core) (*Logger, error) {
```
khushijain21, line 261: `typedMsg` is a log message snippet that we expect to contain sensitive information from ES exporter

`typedMsg` from beatreceivers should be passed as 
`failed to index document; input may contain sensitive data

https://github.com/open-telemetry/opentelemetry-collector-contrib//blob/main/exporter/elasticsearchexporter/bulkindexer.go#L384

This way we can have feature parity with the original `logging.event_data`. All documents rejected by ES exporter, containing sensitive information will be redirected to `beatname-event-data.ndjson` file

mauri870:
I don't see anything out of place, it just missing a test for `ConfigureOTelCore`.

logp/core.go
```diff
@@ -248,12 +243,27 @@ func ConfigureWithCoreLocal(loggerCfg Config, core zapcore.Core) (*Logger, error
...
-		level:        level,
+		level:        zap.NewAtomicLevelAt(loggerCfg.Level.ZapLevel()),
 		observedLogs: nil,
 	})
 	return newLogger(root, ""), nil
 }
 
+// ConfigureEventLoggingOTel takes a list log messages expected to contain sensitive data
+// and ensures they are logged using typed logger
+func ConfigureEventLoggingOTel(typedMsg []string, core zapcore.Core) (*Logger, error) {
```
khushijain21, line 254: typedMsg is a log message snippet that we expect to contain sensitive information 

typedMsg should be passed as `failed to index document; input may contain sensitive data` to log failed docs from ES exporter to `-event.log` file https://github.com/open-telemetry/opentelemetry-collector-contrib//blob/main/exporter/elasticsearchexporter/bulkindexer.go#L384

pchila:
@khushijain21 Could you please explain why a logger package in a library should concern itself with Otel or event logging filtering ? Those are concepts that are specific to the related applications (elastic-agent in this case) so I feel that we are giving the library additional responsibilities that do not belong in a generic logging package

khushijain21:
Elastic-agent already uses `typedLoggerCore` to filter events based on provided key-value pair. 
https://github.com/khushijain21/elastic-agent/blob/eventLogging/pkg/core/logger/logger.go#L133
This `typedLogger` core is further created by the user provided config under `agent.logging.event_data`
[For Ref](https://github.com/khushijain21/elastic-agent/blob/eventLogging/pkg/core/logger/logger.go#L133)

We are only extending that functionality to include certain "log message snippets". If these log message exists - we log them using `typedCore`. This was specifically required for OTel hence the naming of new methods, but I could make them generic.

The `logp` package is concerned here because where and how events are logged comes from user configuration, and we want certain log message (coming from the OTel world) to use the same core. I hope that makes sense

PR #328, opened 2025-07-09
User agent update

graphaelli:
## What does this PR do?

Expands the possibilities for user agent to include unmanaged beats.

It also relaxes the 100 character enforcement introduced recently.

## Checklist

## Related issues

- relates #307

### DIFF
--- a/useragent/useragent.go
+++ b/useragent/useragent.go
@@ -18,7 +18,6 @@
 package useragent
 
 import (
-	"errors"
 	"runtime"
 	"strings"
 )
@@ -90,7 +89,7 @@ func UserAgent(binaryNameCapitalized string, version, commit, buildTime string, 
 	return builder.String()
 }
 
-func UserAgentWithBeatTelemetry(binaryNameCapitalized string, version string, mode AgentManagementMode, unprivileged AgentUnprivilegedMode) (string, error) {
+func UserAgentWithBeatTelemetry(binaryNameCapitalized string, version string, mode AgentManagementMode, unprivileged AgentUnprivilegedMode) string {
 	var builder strings.Builder
 	builder.WriteString("Elastic-" + binaryNameCapitalized + "/" + version + " ")
 	uaValues := []string{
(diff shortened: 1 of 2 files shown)


useragent/useragent.go
```diff
@@ -82,23 +93,22 @@ func UserAgent(binaryNameCapitalized string, version, commit, buildTime string,
...
+	if unprivileged != AgentUnprivilegedModeUnknown {
+		uaValues = append(uaValues, unprivileged.String())
 	}
 	builder.WriteByte('(')
 	builder.WriteString(strings.Join(uaValues, "; "))
 	builder.WriteByte(')')
 
-	// Ensure the user agent string does not exceed 100 characters
-	userAgent := builder.String()
-	if len(userAgent) > 100 {
```
pierrehilbert, line 100: Having this check was a requirement from https://github.com/elastic/ingest-dev/issues/5333 considering our proxy logs are limiting the user agent size to 100.

graphaelli: I'd be ok to add a failing test but it shouldn't be in the API - what's the consumer supposed to do it if it fails?  The previous implementation still returned the calculated user agent but using that when error is not nil is too non-idiomatic.

useragent/useragent.go
```diff
@@ -18,22 +18,27 @@
...
-	AgentManagementModeManaged AgentManagementMode = iota
-	// AgentManagementModeStandalone indicates that the beat is running in standalone mode.
+	// AgentManagementModeUnknown indicates that the management mode is unknown.
+	AgentManagementModeUnknown AgentManagementMode = iota
+	// AgentManagementModeUnmanaged indicates that the beat is not running under agent.
+	AgentManagementModeUnmanaged
+	// AgentManagementModeManaged indicates that the beat is running under agent managed by Fleet.
+	AgentManagementModeManaged
+	// AgentManagementModeStandalone indicates that the beat is running under agent in standalone mode.
 	AgentManagementModeStandalone
```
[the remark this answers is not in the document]
graphaelli: No strong opinion here, whatever is least surprising works for me - @cmacknz can you tie break here?

useragent/useragent.go
graphaelli: added this as a test

PR #330, opened 2025-07-24
[monitoring] Add ClearOnVisitHistogram to adapter for go-metrics

leehinman:
## What does this PR do?

Adds a new type of ClearOnVisitHistogram to monitoring/adapter.  

## Why is it important?

This type is the same as a go-metrics StandardHistogram, but the Visit function for the new type will call `Clear` on the underlying sample.  This means the sample is reset each time you visit it.  This is useful for things like the beats "Non-zero metrics in the last 30s" log entries.

The other difference is that the `Visit` function only reports the following metrics: 

  - count
  - min
  - max
  - median
  - p99

This is smaller than the fields produced for the StandardHistogram.  This was done to limit the field expansion, and this list of fields was picked because it is the list actually used in the elastic-agent monitoring dashboards.

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/monitoring/adapter/go-metrics-wrapper.go
+++ b/monitoring/adapter/go-metrics-wrapper.go
@@ -25,7 +25,7 @@ import (
 
 // go-metrics wrapper interface required to unpack the original metric
 type goMetricsWrapper interface {
-	wrapped() interface{}
+	wrapped() any
 }
 
 // go-metrics wrappers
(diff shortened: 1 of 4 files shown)


mauri870 (approved):
I don't see any obvious issues, but I'm not very experienced with how this integrates with the agent dashboards. I'd give more weight to Fae's review.

leehinman:
@mauri870 & @faec I did a draft PR of how I think this could be used in the beats repo. 

https://github.com/elastic/beats/pull/45749

PR #333, opened 2025-07-29
[Windows] Rotate (rename) files with retries

ycombinator:
## What does this PR do?

This PR enhances the `SafeFileRotate` implementation take a variadic `opts` argument.  This argument can be used to configure the function to retry it's final rename step with a specific interval and up to a specified duration.  On Windows, the default behavior is to retry.

This is similar to a fix made for [Agent uninstallation](https://github.com/elastic/elastic-agent/blob/8fbdaaa979db0976e33612be647bb69ffaac70f8/internal/pkg/agent/install/uninstall.go#L277-L284).

## Why is it important?

Windows does not permit renaming files that are currently being held open by other processes, e.g. antivirus scanners.  Retrying the rename increases the likelihood that the rename will succeed on one of the retry attempts when the file is not being help open by another process.

## Checklist

## How to test this PR locally

1. On a Windows machine, reproduce the failing scenario.

   1. Revert the commit with the fix.
   ```
   git checkout a711c5aa00e500d7136d95e2b658db15e974261a
   ```
   2. Run the unit test introduced in this PR and ensure that it fails with the `Access denied` error.
   ```
   go test -v ./file/... -run ^TestSafeFileRotate$
   ```

2. On a Windows machine, re-run the same test with the fix.
   1. Checkout this PR's branch.
   2. Run the unit test introduced in this PR multiple times and ensure that it no longer fails with the `Access denied` error.
   ```
   go test -v ./file/... -run ^TestSafeFileRotate$ -count 10
   ```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/5862

### DIFF
--- a/file/helper_common.go
+++ b/file/helper_common.go
@@ -6,7 +6,7 @@
 // not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//	http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing,
 // software distributed under the License is distributed on an
--- a/file/helper_common.go
+++ b/file/helper_common.go
@@ -42,6 +42,8 @@ func rename(src, dst string, options RotateOpts) error {
 		return os.Rename(src, dst)
 	}
 
+	// Attempt rename with retries every options.RenameRetryInterval until options.RenameRetryDuration
+	// has elapsed. This is useful in cases where the destination file may be locked or in use.
 	var err error
 	for start := time.Now(); time.Since(start) < options.RenameRetryDuration; time.Sleep(options.RenameRetryInterval) {
 		err = os.Rename(src, dst)
--- a/file/helper_common.go
+++ b/file/helper_common.go
@@ -38,7 +38,8 @@ func WithRenameRetries(duration, interval time.Duration) RotateOpt {
 }
 
 func rename(src, dst string, options rotateOpts) error {
-	if options.renameRetryDuration == 0 && options.renameRetryInterval == 0 {
+	// Perform a regular (non-retrying) rename unless all retry options are specified.
+	if options.renameRetryDuration == 0 || options.renameRetryInterval == 0 {
 		return os.Rename(src, dst)
 	}
 
--- a/file/helper_common.go
+++ b/file/helper_common.go
@@ -1,0 +1,65 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package file
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+type RotateOpts struct {
+	RenameRetryDuration time.Duration
+	RenameRetryInterval time.Duration
+}
+
+type RotateOpt func(*RotateOpts)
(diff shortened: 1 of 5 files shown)


swiatekm:
I feel like this should either be a new function, or we should change the signature and make the retry timings configurable. A file operation potentially blocking for two seconds sounds like it might be unexpected for users, even if the alternative is to get an error on Windows.

ycombinator:
>  Is the intent here to retry if the file happened to be held temporarily by an antivirus or similar tool?

Yes, that's exactly right.  This is why I feel like the retry is an OS (Windows)-specific implementation detail.  The same function's implementation for other OSes does not have retries.

pchila:
> FWIW, we do something similar during Agent uninstallation: https://github.com/elastic/elastic-agent/blob/8fbdaaa979db0976e33612be647bb69ffaac70f8/internal/pkg/agent/install/uninstall.go#L277-L284

TBH during uninstall we don't really care about the contents of the files, we just want to delete them.
The 

> FWIW, we do something similar during Agent uninstallation: https://github.com/elastic/elastic-agent/blob/8fbdaaa979db0976e33612be647bb69ffaac70f8/internal/pkg/agent/install/uninstall.go#L277-L284

To be fair, during uninstall we don't care about the final contents of a file, we just want them gone, here we are writing file content and the conflict may arise from some other concurrent writes and the final result may be unexpected.
Retrying like this may overwrite updates happening on the same file, I would rather give control of retries to the caller  rather than silently brute force through a file lock for 2 seconds.

ycombinator:
> here we are writing file content and the conflict may arise from some other concurrent writes and the final result may be unexpected.
> Retrying like this may overwrite updates happening on the same file, I would rather give control of retries to the caller rather than silently brute forcing our way through a file lock for 2 seconds.

I'm not sure I understand how a conflict may arise.  If another process is writing to this file, this function would not be able to rename it until the other process closes its handle, no?

pchila:
> [EDIT] Come to think of the callback idea — that sounds like overkill as the caller could simply use the function as it is today and implement any additional logic around it on it's end. So I would not go with this idea. Either we just allow the caller to control the retry frequency and duration or we leave this function as it is today (before this PR) and implement the retry logic entirely on the caller's end around the existing function call. WDYT?

I would be in favor of handling the retries at the point where new content is saved to fleet.enc by calling this function.

ycombinator:
> > [EDIT] Come to think of the callback idea — that sounds like overkill as the caller could simply use the function as it is today and implement any additional logic around it on it's end. So I would not go with this idea. Either we just allow the caller to control the retry frequency and duration or we leave this function as it is today (before this PR) and implement the retry logic entirely on the caller's end around the existing function call. WDYT?
> 
> I would be in favor of handling the retries at the point where new content is saved to fleet.enc by calling this function.

Got it, thanks.  I'll close this PR then to leave the implementation of `SafeFileRotate` as it is today and open a new PR against `elastic-agent` to add the retry logic where `fleet.enc` is being saved by calling `SafeFileRotate`.

ycombinator:
Bringing back the implementation in this PR in favor of https://github.com/elastic/elastic-agent/pull/9224. See https://github.com/elastic/elastic-agent/pull/9224#discussion_r2273632334 for details.

file/helper_common.go
```diff
@@ -0,0 +1,61 @@
...
+	return func(opts *RotateOpts) {
+		opts.RenameRetryDuration = duration
+		opts.RenameRetryInterval = interval
+	}
+}
+
+func rename(src, dst string, options RotateOpts) error {
+	if options.RenameRetryDuration == 0 && options.RenameRetryInterval == 0 {
+		return os.Rename(src, dst)
+	}
```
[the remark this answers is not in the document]
ycombinator: Fixed in d36f8df64f2f763a7509f3da3baa07995635be16.

file/helper_windows.go
```diff
@@ -23,9 +23,13 @@ import (
 )
 
 // SafeFileRotate safely rotates an existing file under path and replaces it with the tempfile
-func SafeFileRotate(path, tempfile string) error {
+func SafeFileRotate(path, tempfile string, opts ...RotateOpt) error {
+	options := rotateOpts{}
```
blakerouse, line 27: I think default we should have the `renameRetryDuration` and `renameRetryInterval` set to the defaults on Windows. At the moment you have to use the options to get the behavior, I think having it the default is more preferred on Windows.

blakerouse (changes requested):

file/helper_windows.go
ycombinator: Addressed in 04a4d5e.

PR #345, opened 2025-08-26
feat: remove WithHTTP2Only option

kruskall:
## What does this PR do?

HTTP/2 has been enabled by default since a long time

the x/net/http2 package has been mostly integrated in the stdlib and ForceAttemptHTTP2 is used to control HTTP/2

## Why is it important?

drop x/net dependency usage

verified it's not used anywhere downstream

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/transport/httpcommon/httpcommon.go
+++ b/transport/httpcommon/httpcommon.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	"go.elastic.co/apm/module/apmhttp/v2"
-	"golang.org/x/net/http2"
 
 	"github.com/elastic/elastic-agent-libs/config"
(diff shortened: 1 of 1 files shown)


khushijain21:
> HTTP/2 has been enabled by default since a long time

@kruskall this library force uses http/1 in its default scenario because of this config. See https://github.com/elastic/elastic-agent-libs/blob/89b303faeab570c9353b0dd1979476e42c4285cf/transport/httpcommon/httpcommon.go#L315

So using http/2 is still opt-in

kruskall:
mmh, we might want to use that field instead of the http2 package, I'll rework this

PR #358, opened 2025-09-25
add proxy test

khushijain21:
## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/testing/proxytest/https.go
+++ b/testing/proxytest/https.go
@@ -1,0 +1,203 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+
+package proxytest
+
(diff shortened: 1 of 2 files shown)


AndersonQ (changes requested):
@khushijain21, could you bring the whole `proxytest` package from the agent repo? On [`elastic-agent/testing/proxytest`](https://github.com/elastic/elastic-agent/tree/615190be3f6a28403883f6b4983764558173e0e7/testing/proxytest) there are 2 important files which are not there, the tests for the proxy and an example of how to use it as a "cli" application.

belimawr (commented):
I agree with @AndersonQ, it's better to bring the whole package, which includes the tests.

> It also makes a small change to the original code. proxy-server now writes the incoming headers first and then the status code.

This is actually a bugfix. The headers should be fully set before calling `WriteHeader`. Thanks for fixing that!

khushijain21:
> could you bring the whole proxytest package from the agent repo? 

done

PR #359, opened 2025-09-26
Export tlscommon.TLSCurveType

ycombinator:
⚠️ _Do not merge yet - we may not need this PR after al!_ ⚠️ 

## What does this PR do?

This PR exports the `tlscommon.TLSCurveType` type.

## Why is it important?

So other packages can import it, e.g. https://github.com/elastic/fleet-server/pull/5562


swiatekm (approved):
LGTM, but please fix the linter warnings.

PR #362, opened 2025-10-22
[logp/logptest] Add logptest.Logger

belimawr:
## What does this PR do?

This PR adds logptest.Logger, a logger meant to be used on tests, its key features:
 - All logs are saved on a single temporary log file
 - No logs are printed during test execution
 - On failures, the log file is kept and its path printed
 - Methods to search and wait for log entries are provided, they keep track of the offset, ensuring ordering when when searching for logs

## Why is it important?

It adds a logger to be used in tests with some helper functions

## Checklist

~~## Author's Checklist~~
~~## Related issues~~

### DIFF
--- a/logp/logptest/logger_test.go
+++ b/logp/logptest/logger_test.go
@@ -106,7 +106,7 @@ func assertLogFormat(t *testing.T, path string) {
 		}
 
 		if entry.LogLevel == "" {
-			t.Error("'log.level' cannot be emtpy")
+			t.Error("'log.level' cannot be empty")
 		}
 
 		if entry.Message == "" {


logp/logptest/logger.go
```diff
@@ -0,0 +1,187 @@
...
+			if !found {
+				c.Errorf("did not find '%s' in the logs", s)
+			}
+		},
+		timeout,
+		100*time.Millisecond,
+		msgAndArgs...)
+}
+
+// logContains searches for str in the log file keeping track of the offset.
```
orestisfl:
```suggestion
// LogContains searches for str in the log file keeping track of the offset.
```

logp/logptest/logger.go
```diff
@@ -0,0 +1,187 @@
...
+// Logger wraps a *logp.Logger and makes it more suitable for tests.
+// Key features:
+//   - All logs are saved on a single temporary log file
+//   - On failures, the log file is kept and its path printed
+//   - Methods to search and wait for log entries are provided,
+//     they keep track of the offset, ensuring ordering when
+//     when searching for logs
+type Logger struct {
+	*logp.Logger
+	t       *testing.T
```
orestisfl:
```suggestion
```

I'd prefer if `t` is treated as one of the objects, like `context.Context` that is hierarchical and should not be stored in structs but passed as an argument to functions. This is because it's hierarchical and users of this package may desire to use one logger but multiple `t`s (via `t.Run` subtests).

logp/logptest/logger.go
```diff
@@ -0,0 +1,187 @@
...
+func (l *Logger) LogContains(str string) {
+	l.t.Helper()
+	found, err := l.logContains(str)
+	if err != nil {
+		l.t.Fatalf("cannot read log file: %s", err)
+	}
+
+	if !found {
+		l.t.Fatalf("'%s' not found in logs", str)
+	}
```
orestisfl:
```suggestion
	return assert.Truef(l.t, found, "did not find '%s' in the log", str)
```

this is not really a fatal error, users might desire to continue with the test

orestisfl (commented):
I would also export both `t`-based and non-`t`-based functions for users that do not want to fail their tests for non-matching logs. E.g. we might want to test that logs do NOT contain something

logp/logptest/logger.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+// Logger wraps a *logp.Logger and makes it more suitable for tests.
+// Key features:
+//   - All logs are saved on a single temporary log file
+//   - On failures, the log file is kept and its path printed
+//   - Methods to search and wait for log entries are provided,
+//     they keep track of the offset, ensuring ordering when
+//     when searching for logs
+type Logger struct {
```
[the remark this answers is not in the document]
belimawr: That's a great idea!

logp/logptest/logger.go
AndersonQ: If you aren't failing the test immediately, it should not be called `Require*`, but `Assert*` so it's consistent with we're used to with testify

logp/logptest/logger.go
```diff
@@ -0,0 +1,194 @@
...
+// to search in the logs.
+// If dir is not an empty string, the log file will be generated on
+// this folder. This is useful to make CI collect the logs in case
+// a test fails. If dir is an empty string, the OS temporary folder
+// is used.
+//
+// The *logp.Logger is embedded into it, so [Logger] is a drop-in
+// replacement for a *logp.Logger, or the logger can be accessed via
+// [Logger.Logger]
+func NewFileLogger(t *testing.T, dir string) *Logger {
```
mauri870: Mind adding some testing coverage for the functions defined here? You can probably cover them all in a single test function.

belimawr: I can add some testing, thanks. You made me remember I can use `testing.TB` so I can mock `testing.T` and not have the tested code interfere with the actual test.

mauri870 (approved):

PR #390, opened 2026-03-27
feat(mapstr): add DeepCloneUpdate for single-pass clone+merge

strawgate:
## Summary

Adds `M.DeepCloneUpdate(src M)` and `M.DeepCloneUpdateNoOverwrite(src M)` — new methods that merge a source map into a destination like `DeepUpdate`/`DeepUpdateNoOverwrite`, but create fresh copies of all nested maps instead of aliasing the source's sub-maps.

These are equivalent to `dst.DeepUpdate(src.Clone())` / `dst.DeepUpdateNoOverwrite(src.Clone())` but perform both operations in a single pass, avoiding the intermediate clone allocation.

## Motivation

In the beats processor pipeline, shared processor data (add_fields, add_cloud_metadata, add_host_metadata, etc.) is merged into events on every event. Today each processor clones its shared data then DeepUpdates the clone into the event — two passes over the same map tree:

1. `Clone()` walks the source and creates fresh maps (which are immediately discarded after step 2)
2. `DeepUpdate()` walks the cloned source into the destination

`DeepCloneUpdate` combines these into a single walk that clones-on-write directly into the destination.

## Changes from v1

- Renamed `DeepCopyUpdate` → `DeepCloneUpdate` to match existing `Clone` terminology (per reviewer feedback)
- Reverted `goimports` version pin — no longer needed since `.go-version` was fixed upstream

## How to test

```bash
go test -v -race ./mapstr/
go test -bench=BenchmarkDeepCloneUpdate -benchmem ./mapstr/
```


dev-tools/mage/fmt.go
```diff
@@ -31,7 +31,9 @@ import (
 
 const (
 	// GoImportsImportPath controls the import path used to install goimports.
-	GoImportsImportPath = "golang.org/x/tools/cmd/goimports@latest"
+	// Pinned to v0.42.0 because v0.43.0+ requires Go 1.25 and CI
+	// runners use GOTOOLCHAIN=local with Go 1.24.
+	GoImportsImportPath = "golang.org/x/tools/cmd/goimports@v0.42.0"
```
cmacknz: This is fixed now https://github.com/elastic/elastic-agent-libs/blob/main/.go-version

cmacknz:
Looks reasonable to me besides the version pinning that shouldn't be necessary anymore.

I also think this would be slightly nicer to name the new functions `DeepCloneUpdate` to match the existing terminology but this is a nit.

strawgate:
Addressed both items:

- **Renamed** `DeepCopyUpdate` → `DeepCloneUpdate` (and `DeepCopyUpdateNoOverwrite` → `DeepCloneUpdateNoOverwrite`) to match the existing `Clone` terminology.
- **Reverted** the `goimports` version pin — `.go-version` is fixed upstream now so `@latest` works again.

Also squashed into a single clean commit.

cmacknz (approved):

cmacknz:
> Also squashed into a single clean commit.

Then why are there 3 commits in this PR ;)

strawgate:
> > Also squashed into a single clean commit.
> 
> Then why are there 3 commits in this PR ;)

this was a test and you passed 😛

VihasMakwana (approved):

PR #398, opened 2026-04-02
Remove support for encrypted PKCS#1 PEM keys

### LINKED ISSUE #46980: Drop support for insecure PEM-level encryption for private keys in TLS
**Describe the enhancement:**

Some users might still be storing their TLS private keys in PEM files using the insecure and deprecated PEM-level encryption ([PKCS#1](https://www.rfc-editor.org/rfc/rfc8017) key format, [RFC 1423](https://www.rfc-editor.org/rfc/rfc1423) encryption). This encryption does not support authentication and is vulnerable to padding oracle attacks that can let an attacker recover the plaintext of the private key.

This affects all the `output.*.ssl.key` config parameters: 

https://www.elastic.co/docs/reference/beats/filebeat/configuration-ssl

Although it's a full responsibility of the user to choose a secure encryption for their private keys, on our side we still let to use this insecure encryption in our TLS library:

https://github.com/elastic/elastic-agent-libs/blob/31afddcf5b46d16ea201fab0dfe3225b085707ed/transport/tlscommon/tls.go#L117-L124

Naturally, we also support a secure encryption with authentication for private keys passed in the [PKCS#8](https://www.rfc-editor.org/rfc/rfc5208) format:

https://github.com/elastic/elastic-agent-libs/blob/31afddcf5b46d16ea201fab0dfe3225b085707ed/transport/tlscommon/tls.go#L125-L132

This task is to remove support for **encrypted** private keys in the [PKCS#1](https://www.rfc-editor.org/rfc/rfc8017) format that use the PEM-level encryption defined by [RFC 1423](https://www.rfc-editor.org/rfc/rfc1423).

We should also print a well-readable error message on Beat startup when [PKCS#1](https://www.rfc-editor.org/rfc/rfc8017) format is used for encrypted private keys. The message should clearly explain that the PEM-level encryption available for the [PKCS#1](https://www.rfc-editor.org/rfc/rfc8017) format is insecure and might lead to the content of the private key to be exposed.

The message should also explain that the customer should rather use the [PKCS#8](https://www.rfc-editor.org/rfc/rfc5208) format instead because it supports secure encryption with authentication (defined in [PKCS#5](https://www.rfc-editor.org/rfc/rfc8018)).

**Describe a specific use case for the enhancement or feature:**

Customers might be unaware that they are using insecure encryption for their private keys when setting up TLS in Beats. We should properly notify them and drop support for such insecure encryption.

colecschmidt:
## What does this PR do?

This PR removes support for encrypted PKCS#1 PEM private keys, which rely on deprecated and insecure PEM-level encryption (RFC 1423).

Previously, encrypted PEM blocks were decrypted using `x509.DecryptPEMBlock`. This path has been removed, and encrypted PKCS#1 PEM blocks are now rejected with a clear error message.

Support for encrypted PKCS#8 private keys remains unchanged.

## Why is it important?

PEM-level encryption for PKCS#1 keys is deprecated and considered insecure. Removing support prevents use of unsafe cryptographic practices and aligns with modern standards.

## Changes

- Removed PKCS#1 PEM decryption (`decryptPKCS1Key`)
- Replaced decryption path with explicit error
- Kept PKCS#8 encrypted key support intact
- Updated tests to reflect new behavior

## Related issues

Closes #46980


colecschmidt:
Happy to make any changes or adjustments if needed. Thanks for reviewing!

cmacknz:
I'm in favour of removing this, but it is a breaking change and we have no way to know how many people might be relying on this with or without knowing it.

One way we can handle this is in a less disruptive way is to introduce configuration to switch between treating use of PKCS#1 PEM keys as an error or allowing it but logging a warning.

At the package level here, we need to default it to logging the warning. In the consuming products (beats, elastic-agent, fleet-server, etc) we can change the configuration default to consider this an error on main which represents the next minor release. We would leave this as only a warning when this is back-ported into existing release branches so that users won't have to deal with this in a patch release. Ideally our ssl configuration block would allow switching from error to warning only as well to give users control of what happens before we just remove this.

Then in the next minor if there is no feedback from the breaking change we can remove the code entirely.

colecschmidt:
Thanks for the detailed feedback @cmacknz — that's a really good point about backwards compatibility. I hadn't considered the impact on users who might be relying on PKCS#1 without realizing it.

I'm happy to take the approach you described: introduce a warning mode first rather than a hard error, with configuration to allow switching between the two. Before I start, a couple of quick questions to make sure I get the implementation right:

1. For the package-level default (warning vs error), should the new config option live in the existing TLSConfig struct, or somewhere else?
2. Is there a pattern elsewhere in the codebase I should follow for this kind of deprecation warning logging?

Happy to make whatever changes are needed — just want to make sure I'm aligned with how you'd like to see it structured before I dive in.

cmacknz:
In the context of this PR we should:

1. The PCKS PEM decryption should be updated to always log a warning about deprecation https://github.com/elastic/elastic-agent-libs/blob/54d156930579246331b87efc8d0a1111861d6f48/transport/tlscommon/tls.go#L117-L118

2. ReadPEMFile should be updated to accept a configuration flag to treat decrypting PCKS PEM as an error. The default must be to log a warning. https://github.com/elastic/elastic-agent-libs/blob/54d156930579246331b87efc8d0a1111861d6f48/transport/tlscommon/tls.go#L89

3. This configuration flag should be added to CertificateConfig so it can be controlled in our ssl configuration blocks. Let's call it something like `DisablePEMFileSupport` or something similar https://github.com/elastic/elastic-agent-libs/blob/54d156930579246331b87efc8d0a1111861d6f48/transport/tlscommon/types.go#L378-L384

This will position us to be able to disable this support in a gradual way. First we warn about it, then we change the default to error with the option to fall back, then we can remove support completely.

colecschmidt:
@cmacknz I've implemented the warning mode as discussed. The default now logs a deprecation warning and decrypts the key, with disable_legacy_pem_support: true available to opt into the error behavior. All tests pass. Please take another look when you get a chance — happy to make any adjustments.

transport/tlscommon/tls.go
```diff
@@ -114,14 +116,21 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
 		}
 
 		switch {
-		case x509.IsEncryptedPEMBlock(block): //nolint: staticcheck // deprecated, we have to get rid of it
-			block, err := decryptPKCS1Key(*block, pass)
-			if err != nil {
+		case x509.IsEncryptedPEMBlock(block): //nolint: staticcheck // deprecated PKCS#1 PEM encryption
+			if disableLegacy {
+				err = fmt.Errorf("encrypted PKCS#1 PEM keys are insecure and no longer supported; please convert the key to encrypted PKCS#8")
```
cmacknz: You should probably early return out of this block, otherwise it looks like you might return a chain of this same error for every block with a top level `no PEM blocks` message if you couldn't decode any of them.

transport/tlscommon/decrypt_block_nofips.go
```diff
@@ -34,19 +34,12 @@ func decryptPKCS1Key(block pem.Block, passphrase []byte) (pem.Block, error) {
...
+		return block, fmt.Errorf("failed to decrypt PKCS#1 key: %w", err)
 	}
-
-	// DEK-Info contains encryption info. Remove header to mark block as
-	// unencrypted.
-	delete(block.Headers, "DEK-Info")
-	block.Bytes = buffer
-
+	block.Bytes = decrypted
+	block.Headers = nil
```
cmacknz, line 46: This is only equivalent if the only header were `DEK-Info`, is that always the case?

cmacknz (commented):
This follows what I asked for, thanks! I think the error message when disable legacy support is true might be unnecessarily long though in practice and we should make it more concise for users that will see this.

transport/tlscommon/decrypt_block_nofips.go
cmacknz: If the comment is correct that `Proc-Type` also needs to be removed why did removing only `DEK-Info` work before?

colecschmidt:
@cla-checker-service recheck

transport/tlscommon/tls.go
colecschmidt: This is already handled in the refactoring commit — the block now returns early with the error.

transport/tlscommon/decrypt_block_nofips.go
colecschmidt: `x509.IsEncryptedPEMBlock` only checks for `DEK-Info`, so removing just that header was enough to prevent re-detection. However, `Proc-Type` would still falsely appear as `4,ENCRYPTED` in re-encoded PEM output if left in place. We nil the whole map to remove both. Updated the comment to explain this.

transport/tlscommon/tls.go
```diff
@@ -114,14 +120,18 @@ func ReadPEMFile(log *logp.Logger, s, passphrase string) ([]byte, error) {
...
-			block, err := decryptPKCS1Key(*block, pass)
-			if err != nil {
-				log.Errorf("Dropping encrypted pem block with private key, block type '%s': %s", block.Type, err)
-				errs = errors.Join(errs, err)
+		case x509.IsEncryptedPEMBlock(block): //nolint: staticcheck // deprecated PKCS#1 PEM encryption
+			if disableLegacy {
+				return nil, fmt.Errorf("encrypted PKCS#1 PEM keys are not supported; convert to PKCS#8")
+			}
+			log.Warnf("Encrypted PKCS#1 PEM keys are deprecated and insecure, support will be removed in a future release; please convert the key to encrypted PKCS#8")
+			decBlock, decErr := decryptPKCS1Key(*block, pass)
```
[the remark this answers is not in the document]
rdner: That's intended.

transport/tlscommon/tls_nofips_test.go
```diff
@@ -20,114 +20,189 @@
...
-		_, err := LoadTLSConfig(mustLoad(t, fmt.Sprintf(`
+    t.Run("passphrase value", func(t *testing.T) {
+        cfg, err := LoadTLSConfig(mustLoad(t, `
+        enabled: true
+        certificate: testdata/ca.crt
+        key: testdata/ca.encrypted.key
+        key_passphrase: Abcd1234!
+        `), logptest.NewTestingLogger(t, ""))
+        require.NoError(t, err)
+        assert.NotNil(t, cfg)
```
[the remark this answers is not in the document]
rdner: This looks like a valid concern to me.

rdner:
@colecschmidt thanks a lot for addressing the feedback.

I see some strange formatting issues in `transport/tlscommon/tls_nofips_test.go`.

The linter issue reported in `transport/tlscommon/tls.go` has nothing to do with the changes in this PR and can be ignored.

Also, the LLM feedback on the test here https://github.com/elastic/elastic-agent-libs/pull/398#discussion_r3160459068 makes sense to me.

The rest looks great. Once these 2 issues are addressed, we can merge this.

transport/tlscommon/tls_nofips_test.go
```diff
@@ -20,114 +20,189 @@
...
+    "testing"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
+    "github.com/stretchr/testify/assert"
+    "github.com/stretchr/testify/require"
 
-	"github.com/elastic/elastic-agent-libs/logp/logptest"
+    "github.com/elastic/elastic-agent-libs/logp/logptest"
 )
```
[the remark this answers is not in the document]
colecschmidt: @copilot apply changes based on [this feedback](https://github.com/elastic/elastic-agent-libs/pull/398#discussion_r3160459128)

rdner (approved):
Looks good now, I fixed the formatting issues.

PR #400, opened 2026-04-10
perf: implement go-structform Folder interface on mapstr.M

strawgate:
**Eliminates all allocations from `mapstr.M` JSON encoding via go-structform.**

## Problem

`mapstr.M` is a named type (`type M map[string]interface{}`). Go's type switch treats it as distinct from `map[string]interface{}`, so go-structform's fast path in `getFoldGoTypes` misses it. Every nested `mapstr.M` value falls through to `getFoldConvert` which calls `reflect.Convert` — allocating per map.

In production filebeat profiles (throughput-ea, 30s sample), JSON encoding (via go-structform) is **21% of total CPU**. The `reflect.Convert` + type-switch overhead for named map types is a significant contributor, and the Fold `default` branch creates a new `gotype.Iterator` + `typeFoldRegistry` for every unhandled value — **8.3M + 8.2M allocations** per sample.

## Solution

Implement `gotype.Folder` on `mapstr.M`. go-structform checks for this interface before reaching the reflection path. The `Fold` method:

- Walks the map directly via the structform Visitor API
- Handles all common value types without reflection: `string`, `int`, `int64`, `float64`, `bool`, `nil`, `M`, `map[string]interface{}`, `[]interface{}`
- Falls back to `gotype.Fold` for any other type, guaranteeing byte-identical output
- Includes `foldSlice` helper for `[]interface{}` arrays with the same type coverage

## Benchmark results

Encoding realistic events (post-normalization types: `int64`, `float64`, `bool`) across different event shapes:

| Event shape | Fields | Nested maps | Before ns/op | After ns/op | **Δ time** | Before allocs | After allocs | **Δ allocs** |
|---|---|---|---|---|---|---|---|---|
| flat_5fields | 5 | 0 | 535 | 311 | **-42%** | 12 | 0 | **-100%** |
| nested_10fields | 10 | 3 | 938 | 800 | **-15%** | 8 | 0 | **-100%** |
| nested_20fields | 20 | 7 | 2,278 | 1,930 | **-15%** | 24 | 0 | **-100%** |
| numeric_heavy | 12 | 1 | 1,517 | 706 | **-53%** | 44 | 0 | **-100%** |
| strings_only | 5 | 2 | 477 | 484 | ~0% | 0 | 0 | 0 |

- **All allocations eliminated** for events containing `int64`, `float64`, `bool`, and `[]interface{}` values
- No regression for string-only events (already handled)
- Larger wins on numeric-heavy events where more values fell through to reflect
- `strings_only` confirms zero overhead when all types were already handled

### Production profile impact (estimated)

From the throughput-ea pprof (30s, ~1M events):
- Eliminates ~8.3M `gotype.NewIterator` objects (213 MB)
- Eliminates ~8.2M `gotype.newTypeFoldRegistry` objects (225 MB)
- Eliminates ~9.2M `reflect.New` objects (140 MB)
- **Total: ~578 MB alloc_space, ~25.7M alloc_objects (~19% of total)**

## How to test

```bash
go test -v -race ./mapstr/ -run TestFold
go test -bench BenchmarkFold -benchmem ./mapstr/
```

The `TestFoldMatchesReflection` suite verifies byte-identical JSON output against go-structform's reflection path for 17 event shapes including int64 boundaries, float64 edge cases, nested slices, nil values, and realistic post-normalization events.

### DIFF
--- a/mapstr/fold.go
+++ b/mapstr/fold.go
@@ -72,6 +72,13 @@ func foldSlice(v structform.ExtVisitor, s []interface{}) error {
 // Fold implements go-structform's gotype.Folder interface, letting
 // go-structform serialize M without the reflect.Convert overhead it
 // normally incurs for named map types.
+//
+// The type switch handles common leaf types (string, int, int64, float64,
+// bool, nil) directly rather than delegating to gotype.Fold. This is
+// intentional: each gotype.Fold call allocates an iterator and type
+// registry (~28 B), so falling back for every leaf value adds 20-72
+// allocs/op depending on event shape (2x slower on typical events).
+// Uncommon types still fall through to gotype.Fold for full coverage.
 func (m M) Fold(v structform.ExtVisitor) error {
 	if err := v.OnObjectStart(len(m), structform.AnyType); err != nil {
 		return err
--- a/mapstr/fold.go
+++ b/mapstr/fold.go
@@ -1,0 +1,60 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package mapstr
+
+import (
+	structform "github.com/elastic/go-structform"
+	"github.com/elastic/go-structform/gotype"
+)
+
+// Fold implements go-structform's gotype.Folder interface, letting
+// go-structform serialize M without the reflect.Convert overhead it
+// normally incurs for named map types.
+func (m M) Fold(v structform.ExtVisitor) error {
+	if err := v.OnObjectStart(len(m), structform.AnyType); err != nil {
+		return err
+	}
+	for k, val := range m {
+		if err := v.OnKey(k); err != nil {
+			return err
+		}
+		switch x := val.(type) {
+		case string:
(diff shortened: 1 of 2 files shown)


mapstr/fold.go
```diff
@@ -0,0 +1,127 @@
...
+				return err
+			}
+		case bool:
+			if err := v.OnBool(x); err != nil {
+				return err
+			}
+		case nil:
+			if err := v.OnNil(); err != nil {
+				return err
+			}
```
VihasMakwana, line 54: Could we let `gotype.Fold` to handle these types? Looking at gotype, it looks like it can handle these and a few others. https://github.com/elastic/go-structform/blob/4118395bd29965d22459694fb6b3296e18036378/gotype/fold.go#L145

strawgate: Fallback to gotype.Fold results in the allocation of an iterator and a registry which kills the allocs/perf win. So we only fold for uncommon types, this gives us the perf win in the common case with full coverage provided by fallback

cmacknz: Can this explanation be added as a comment?

strawgate: Benchmarked delegating leaf types to `gotype.Fold` vs the current type switch (Apple M2 Pro, 6 runs):                                                             
                                                                                                                                                                  
| Event shape | type_switch ns/op | gotype.Fold ns/op | Δ speed | type_switch allocs | gotype.Fold allocs |                                                       
|---|---|---|---|---|---|                                                                                                                                         
| flat_5fields | 303 | 678 | +124% slower | 0 | 20 |                                                                                                              
| nested_10fields | 790 | 1,376 | +74% slower | 0 | 32 |                                                                                                          
| nested_20fields | 1,808 | 3,196 | +77% slower | 0 | 72 |
| numeric_heavy | 698 | 1,502 | +115% slower | 0 | 44 |                                                                                                           
                                                                                                                                                                
Each `gotype.Fold` call for a leaf value allocates an iterator + registry (~28 bytes), so the allocation count scales linearly with the number of leaf fields in the event. The type switch avoids this entirely.

strawgate: ill add a comment to this part of the code explaining why we arent just using the fallback for everything other than the mapstr

cmacknz (approved):

mapstr/fold.go
VihasMakwana: Thanks!

VihasMakwana (approved):

PR #401, opened 2026-04-15
fix(logp): avoid unconditional fmt.Sprintf in MakeDebug

orestisfl:
## What does this PR do?

    fix(logp): avoid unconditional fmt.Sprintf in MakeDebug

    MakeDebug called fmt.Sprintf on every invocation regardless of log
    level. Add a debug-level check so the formatting is skipped when debug
    logging is disabled, consistent with the other helpers in global.go.

Note: `.Sugar()` would make the change a 1-liner but it would introduce a logger clone 

## Why is it important?

fmt.Sprintf is wasteful and allocates strings that are immediately thrown away.

## Checklist

### DIFF
--- a/logp/global.go
+++ b/logp/global.go
@@ -29,7 +29,10 @@ import (
 // Deprecated: Use logp.NewLogger.
 func MakeDebug(selector string) func(string, ...interface{}) {
 	return func(format string, v ...interface{}) {
-		globalLogger().Named(selector).Debug(fmt.Sprintf(format, v...))
+		log := globalLogger().Named(selector)
+		if log.Core().Enabled(zap.DebugLevel) {
+			log.Debug(fmt.Sprintf(format, v...))
+		}
 	}
 }
 


logp/global.go
```diff
@@ -29,7 +29,10 @@ import (
 // Deprecated: Use logp.NewLogger.
 func MakeDebug(selector string) func(string, ...interface{}) {
```
khushijain21, line 30: Are we using this method anywhere? I mean this is deprecated so even if we are - we should change where it is invoked

orestisfl: Unfortunately, yes https://github.com/search?q=repo%3Aelastic%2Fbeats%20MakeDebug&type=code 19 instances that lead to hundred actual invocations of the callback.

We can refactor to get rid of the deprecated method but until then, there's no reason to suffer the performance cost

orestisfl:
Example of benchmark in beats:
```diff
diff --git a/packetbeat/protos/mongodb/mongodb_test.go b/packetbeat/protos/mongodb/mongodb_test.go
index 5e128178ab..d1b95d04f9 100644
--- a/packetbeat/protos/mongodb/mongodb_test.go
+++ b/packetbeat/protos/mongodb/mongodb_test.go
@@ -369,6 +369,60 @@ func TestOpCodeNames(t *testing.T) {
 	}
 }
 
+func BenchmarkMongodbSimpleFind(b *testing.B) {
+	reqData, _ := hex.DecodeString(
+		"320000000a000000ffffffffd4070000" +
+			"00000000746573742e72667374617572" +
+			"616e7473000000000001000000050000" +
+			"0000")
+	respData, _ := hex.DecodeString(
+		"020200004a0000000a00000001000000" +
+			"08000000000000000000000000000000" +
+			"01000000de010000075f696400558beb" +
+			"b45f075665d2ae862703616464726573" +
+			"730069000000026275696c64696e6700" +
+			"05000000313030370004636f6f726400" +
+			"1b000000013000e6762ff7c97652c001" +
+			"3100d5b14ae9996c4440000273747265" +
+			"657400100000004d6f72726973205061" +
+			"726b2041766500027a6970636f646500" +
+			"060000003130343632000002626f726f" +
+			"756768000600000042726f6e78000263" +
+			"756973696e65000700000042616b6572" +
+			"79000467726164657300eb0000000330" +
+			"002b00000009646174650000703d8544" +
+			"01000002677261646500020000004100" +
+			"1073636f72650002000000000331002b" +
+			"0000000964617465000044510a410100" +
+			"00026772616465000200000041001073" +
+			"636f72650006000000000332002b0000" +
+			"00096461746500009cda693c01000002" +
+			"6772616465000200000041001073636f" +
+			"7265000a000000000333002b00000009" +
+			"646174650000ccb8cd33010000026772" +
+			"616465000200000041001073636f7265" +
+			"0009000000000334002b000000096461" +
+			"7465000014109d2e0100000267726164" +
+			"65000200000042001073636f7265000e" +
+			"0000000000026e616d6500160000004d" +
+			"6f72726973205061726b2042616b6520" +
+			"53686f70000272657374617572616e74" +
+			"5f696400090000003330303735343435" +
+			"0000")
+
+	_, mongodb := mongodbModForTests()
+	tcptuple := testTCPTuple()
+	req := protos.Packet{Payload: reqData}
+	resp := protos.Packet{Payload: respData}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		private := protos.ProtocolData(new(mongodbConnectionData))
+		private = mongodb.Parse(&req, tcptuple, 0, private)
+		mongodb.Parse(&resp, tcptuple, 1, private)
+	}
+}
+
 // Test for a (recovered) panic parsing document length in request/response messages
 func TestDocumentLengthBoundsChecked(t *testing.T) {
 	logp.TestingSetup(logp.WithSelectors("mongodb", "mongodbdetailed"))
```

result:
```text
goos: linux
goarch: amd64
pkg: github.com/elastic/beats/v7/packetbeat/protos/mongodb
cpu: 12th Gen Intel(R) Core(TM) i7-1265U
                    │ /tmp/mongo_baseline.txt │         /tmp/mongo_fix.txt          │
                    │         sec/op          │   sec/op     vs base                │
MongodbSimpleFind-8               15.72µ ± 3%   14.06µ ± 4%  -10.56% (p=0.000 n=10)

                    │ /tmp/mongo_baseline.txt │         /tmp/mongo_fix.txt          │
                    │          B/op           │     B/op      vs base               │
MongodbSimpleFind-8              13.30Ki ± 0%   12.74Ki ± 0%  -4.25% (p=0.000 n=10)

                    │ /tmp/mongo_baseline.txt │        /tmp/mongo_fix.txt         │
                    │        allocs/op        │ allocs/op   vs base               │
MongodbSimpleFind-8                268.0 ± 0%   251.0 ± 0%  -6.34% (p=0.000 n=10)
```

cmacknz (approved):

mauri870 (approved):

faec (approved):

PR #403, opened 2026-04-17
fix(mapstr): DeepCloneUpdate must merge into map[string]interface{} destinations

strawgate:
## Summary

- **Bug fix**: `DeepCloneUpdate` and `DeepCloneUpdateNoOverwrite` only recursed when the destination value was `mapstr.M`. When the destination held a `map[string]interface{}` (common for events decoded from JSON/wire), they overwrote the entire subtree with a fresh map, silently dropping all existing fields.
- **Impact**: This caused broad regressions in beats (elastic/beats#49762): `event.start`, `event.end`, `process.start`, and `@timestamp` were lost or replaced with `{}` whenever a processor merged metadata into events whose field trees contained `map[string]interface{}` nodes.
- **Fix**: Added an explicit `map[string]interface{}` destination branch that casts to `M` and recurses in-place, matching the behavior of `deepUpdateValue`.
- **Tests**: Regression tests for mixed-type destinations (the failing scenario), plus the deep-nested partial-population and sub-map aliasing cases from the original review.

## Test plan

- [x] `go test ./mapstr/... -race` — all pass

### DIFF
--- a/mapstr/deepcloneupdate_test.go
+++ b/mapstr/deepcloneupdate_test.go
@@ -589,6 +589,67 @@ func BenchmarkHeavyPipeline(b *testing.B) {
 	})
 }
 
+// BenchmarkMixedTypePipeline benchmarks merging into an event whose field tree
+// contains map[string]interface{} subtrees — the common case when events are
+// decoded from JSON before processors run. This exercises the map[string]interface{}
+// destination branch added to fix the v0.36.0 regression.
+func BenchmarkMixedTypePipeline(b *testing.B) {
+	// Processor metadata: pure mapstr.M (as built by add_fields, add_cloud_metadata, etc.)
+	sharedMeta := []M{
+		{"elastic_agent": M{"id": "agent-uuid", "snapshot": false, "version": "8.12.0"}},
+		{"agent": M{"id": "agent-uuid"}},
+		{"data_stream": M{"type": "logs", "dataset": "system.syslog", "namespace": "default"}},
+		{"event": M{"dataset": "system.syslog"}},
+		{"cloud": M{
+			"provider": "aws", "region": "us-east-1",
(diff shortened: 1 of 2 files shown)


strawgate:
@AndersonQ would you be able to review?

mapstr/deepcloneupdate_test.go
```diff
@@ -404,6 +589,67 @@ func BenchmarkHeavyPipeline(b *testing.B) {
...
+			"host": map[string]interface{}{
+				"name": "prod-server-01",
+				"os":   map[string]interface{}{"type": "linux", "version": "22.04"},
+			},
+		}
+	}
+
+	b.Run("clone_and_deep_update", func(b *testing.B) {
+		b.ReportAllocs()
+		for i := 0; i < b.N; i++ {
```
belimawr, line 632: [Suggestion] Use `b.Loop()`
```suggestion
		for b.Loop() {
```

mapstr/deepcloneupdate_test.go
```diff
@@ -404,6 +589,67 @@ func BenchmarkHeavyPipeline(b *testing.B) {
...
+			for _, src := range sharedMeta {
+				dst.DeepUpdate(src.Clone())
+			}
+			benchSinkM = dst
+		}
+	})
+
+	b.Run("deep_clone_update", func(b *testing.B) {
+		b.ReportAllocs()
+		for i := 0; i < b.N; i++ {
```
belimawr, line 643: [Suggestion] Use `b.Loop()`
```suggestion
		for b.Loop() {
```

belimawr (approved):

strawgate:
Closing in favour of #413 which addresses the maintainer benchmark feedback (b.Loop()) and is opened from my fork so review/push cycles work properly.

PR #408, opened 2026-04-23
Add retries to the Kibana client

swiatekm:
## What does this PR do?

Adds retries to the Kibana client. 

The configuration is identical to what elasticsearch-go does: https://www.elastic.co/docs/reference/elasticsearch/clients/go/configuration#_config_retry, with the exception that disabling retries can be done by just setting `max_retries` to 0. 

The implementation is very similar to https://github.com/elastic/elastic-transport-go/blob/63cfa05e956c82203e86043bbf06eeab8f0a647f/elastictransport/elastictransport.go#L354.

## Why is it important?

Retries are useful to have. In particular, we use this client in elastic-agent integration tests, and have recently been having trouble with ESS Kibana returning 503s.

## Checklist

### DIFF
--- a/kibana/client.go
+++ b/kibana/client.go
@@ -380,7 +380,7 @@ func addHeaders(out, in http.Header) {
 
 // Implements RoundTrip interface
 func (conn *Connection) RoundTrip(r *http.Request) (*http.Response, error) {
-	return conn.HTTP.Do(r)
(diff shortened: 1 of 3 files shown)


blakerouse (approved):
Nice!

mauri870 (approved):

PR #411, opened 2026-04-24
Revert "perf: implement go-structform Folder interface on mapstr.M"

michalpristas:
Reverts elastic/elastic-agent-libs#400

This one is breaking and prevents updating elastic-agent-libs in beats.
Probably some weird behaviors when working with time.Time. Timeboxed this to an hour did not fix it.

Example of failures
https://buildkite.com/elastic/beats-xpack-auditbeat/builds/27579#019dbf7c-e258-471e-b4be-5d8b66ee53c3

WIth latest libs but this one reverted: 
https://buildkite.com/elastic/beats-xpack-auditbeat/builds/27589


cmacknz (approved):

cmacknz:
This is one reason why I regret separating libs from Beats, the test coverage here is not enough to prove things actually work in practice.

PR #413, opened 2026-04-28
fix(mapstr): DeepCloneUpdate must merge into map[string]interface{} destinations

strawgate:
## Summary

- **Bug fix**: `DeepCloneUpdate` and `DeepCloneUpdateNoOverwrite` only recursed when the destination value was `mapstr.M`. When the destination held a `map[string]interface{}` (common for events decoded from JSON/wire), they overwrote the entire subtree with a fresh map, silently dropping all existing fields.
- **Impact**: This caused broad regressions in beats (elastic/beats#49762): `event.start`, `event.end`, `process.start`, and `@timestamp` were lost or replaced with `{}` whenever a processor merged metadata into events whose field trees contained `map[string]interface{}` nodes.
- **Fix**: Added an explicit `map[string]interface{}` destination branch that casts to `M` and recurses in-place, matching the behavior of `deepUpdateValue`.
- **Tests**: Regression tests for mixed-type destinations (the failing scenario), plus the deep-nested partial-population and sub-map aliasing cases from the original review.

## Test plan

- [x] `go test ./mapstr/... -race` — all pass

## Related Issues:

 - Supersedes #403 (addresses maintainer feedback: benchmarks now use `b.Loop()`).
 - Blocks https://github.com/elastic/beats/pull/49762


strawgate:
@andersonq i had some permissions issue with the previous branch/pr and have re-opened this one with your feedback applied and resolving the macroscope feedback

andrzej-stencel (dismissed):

mauri870 (dismissed):

rdner (approved):

PR #416, opened 2026-04-30
Add redact.Redact method

michel-laterman:
## What does this PR do?

Add a `redact` package with a public `Redact` method to scan through a passed map and redact any sensitive values in place.

Redaction logic can be changed with optional args (i.e., ignore_keys, redaction prefix marker support).

## Why is it important?

Redaction logic occurs in the elastic-agent as a part of diagnostics collection, and in fleet-server when otel clients send effective config. We need a single way to ensure that sensitive values that can appear do not get leaked.

## Checklist

## Follow up work

- Change uses of `diagnostics.Redact` in elastic-agent
- Change redaction in OpAPM handler of fleet-server.

## Related issues

- Relates #415

### DIFF
--- a/redact/redact.go
+++ b/redact/redact.go
@@ -20,7 +20,6 @@ package redact
 import (
 	"fmt"
 	"io"
-	"maps"
 	"net/url"
 	"reflect"
 	"slices"
@@ -90,7 +89,7 @@ func redactMap[K comparable](obj map[K]any, ro *redactOptions) {
 		if keyString, ok := any(key).(string); ok && strings.ToLower(keyString) == "name" {
 			keyVal, ok := val.(string)
 			if ok && redactKey(keyVal, ro) {
-				for vk := range maps.Keys(obj) {
+				for vk := range obj {
 					if vs, ok := any(vk).(string); ok && strings.ToLower(vs) == "value" {
 						obj[vk] = REDACTED
 						break
@@ -136,6 +135,7 @@ func redactMap[K comparable](obj map[K]any, ro *redactOptions) {
 					if ro.markerPrefix != "" && strings.HasPrefix(keyString, ro.markerPrefix) {
 						markers = append(markers, keyString)
 						delete(obj, key)
+						continue
 					}
 				}
 			default:
--- a/redact/redact.go
+++ b/redact/redact.go
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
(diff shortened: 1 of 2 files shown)


ycombinator:
> Closes https://github.com/elastic/elastic-agent-libs/issues/415

Is this PR sufficient to close the issue?  Don't we need corresponding or follow up PRs to elastic-agent and fleet-server too?

redact/redact.go
```diff
@@ -0,0 +1,193 @@
...
+
+	markers := make([]string, 0)
+	for key, val := range obj {
+		// detect if the obj has entries of the form:
+		// - name: Authorization
+		//   value: Bearer SecretValue
+		if keyString, ok := any(key).(string); ok && strings.ToLower(keyString) == "name" {
+			keyVal, ok := val.(string)
+			if ok && redactKey(keyVal, ro) {
+				for vk := range maps.Keys(obj) {
```
ycombinator: Any specific reason to use the `maps.Keys` iterator here instead of the usual `for vk := range obj`?

redact/redact.go
```diff
@@ -0,0 +1,193 @@
...
+					val = REDACTED
+				} else if redactedValue, redact := redactURL(cast); redact {
+					val = redactedValue
+				}
+			case bool: // redaction marker values are always going to be bool, process redaction markers in this case
+				if keyString, ok := any(key).(string); ok {
+					// Find siblings that have the redaction marker.
+					if ro.markerPrefix != "" && strings.HasPrefix(keyString, ro.markerPrefix) {
+						markers = append(markers, keyString)
+						delete(obj, key)
```
ycombinator, line 137: I think we need a `continue` after this line?  Otherwise on line 149 below we will write `obj[key] = val` again?

michel-laterman: Good catch! I'll add it, but I think the redaction marker loop below would also delete this (which is why the unit tests for these are not failing)

ycombinator (dismissed):

andrzej-stencel (dismissed):

redact/redact.go
```diff
@@ -0,0 +1,193 @@
...
+// RedactOption is an optional arg to control how the Redact method works.
+type RedactOption func(ro *redactOptions)
+
+type redactOptions struct {
+	errOut       io.Writer
+	markerPrefix string
+	ignoreKeys   []string
+}
+
+// WithErrorOutput determines where any error messagess that are encountered are written to.
```
mauri870:
```suggestion
// WithErrorOutput determines where any error messages that are encountered are written to.
```

mauri870 (dismissed):

ycombinator (approved):

PR #404, opened 2026-04-17
feat(tlscommon): add CertReloader for TLS certificate hot-reload

ycombinator:
## What does this PR do?

Adds a `CertReloader` type to the `tlscommon` package that periodically reloads TLS certificate and key files from disk, enabling TLS certificate hot-reload without requiring a process restart.

The design follows the [OpenTelemetry Collector's `configtls` approach](https://github.com/open-telemetry/opentelemetry-collector/blob/7201768d04fb628e65cb4e3d5ead69d17c43d59d/config/configtls/configtls.go#L152): on each `GetCertificate` call (i.e., each TLS handshake), it checks whether the reload interval has elapsed and, if so, re-reads the files from disk. No file watchers or extra goroutines needed.

**API:**
- `NewCertReloader(certPath, keyPath string, opts ...CertReloaderOption) (*CertReloader, error)` — validates and loads the initial cert/key pair
- `WithReloadInterval(d time.Duration) CertReloaderOption` — configures how often to re-read files (default: 5s)
- `(*CertReloader).GetCertificate(*tls.ClientHelloInfo) (*tls.Certificate, error)` — for use with `tls.Config.GetCertificate`

**Behavior:**
- Invalid cert/key pairs during reload are silently skipped; the last successfully loaded certificate continues to be served
- Concurrent access is safe via `sync.RWMutex` with double-checked locking to avoid redundant reloads

## Why is it important?

Fleet Server needs TLS certificate hot-reload ([elastic/fleet-server#6433](https://github.com/elastic/fleet-server/issues/6433)) for environments with frequent certificate rotation — Kubernetes (where Secrets are remounted into pods), serverless (where cert-manager rotates certs), and general ops. The team discussed implementing this in `elastic-agent-libs` so it can be reused by other projects that need the same capability.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works

## Author's Checklist

- [x] Tests cover: valid/invalid/missing cert pairs, reload after interval, invalid cert keeps old, no reload before interval

## Related issues

- Relates https://github.com/elastic/fleet-server/issues/6433
- Relates https://github.com/elastic/fleet-server/pull/6838

PR #417, opened 2026-04-30
feat(tlscommon): wire CertReloader into LoadTLSConfig and LoadTLSServerConfig

ycombinator:
## What does this PR do?

Wires `CertReloader` into the core TLS configuration loading functions so that **all consumers** of `tlscommon` get automatic TLS certificate hot-reload without manual wiring.

This is a follow-up to the server-side reloading changes made in #404 and #405. Those PRs added `CertReloader` and the `CertificateReload` config to `ServerConfig`. This PR completes the picture by:

1. **Adding `CertificateReload` to the client-side `Config` struct** — enables cert reload for TLS clients (e.g., Elastic Agent connecting to Fleet Server or Elasticsearch)
2. **Wiring `CertReloader` into `LoadTLSConfig` and `LoadTLSServerConfig`** — when certificate reload is enabled (the default), these functions create a `CertReloader` instead of loading certificates statically
3. **Wiring callbacks in `ToConfig()`** — sets `GetCertificate` and `GetClientCertificate` on the resulting `tls.Config`, clearing `Certificates` so Go's TLS stack uses the callbacks
4. **Adding `GetClientCertificate` method to `CertReloader`** — same reload logic as `GetCertificate`, with the signature expected by `tls.Config.GetClientCertificate`
5. **Adding `resolvePassphrase` helper on `CertificateConfig`** — consolidates passphrase resolution (inline value or file path) used when constructing a `CertReloader`

### Behavior

- **Enabled by default**: certificate reload is on unless explicitly disabled via `certificate_reload.enabled: false`, matching the OTel Collector pattern
- **No extra goroutines**: on each TLS handshake, checks if the reload interval (default 5s) has elapsed; if so, re-reads cert/key files from disk
- **Safe rollover**: invalid cert/key pairs during reload are silently skipped; the last successfully loaded certificate continues to be served
- **Passphrase support**: encrypted private keys work with both `key_passphrase` and `key_passphrase_path`

### Impact on consumers

Consumers that call `LoadTLSConfig` or `LoadTLSServerConfig` (Fleet Server, Elastic Agent, Beats inputs like `http_endpoint`, etc.) automatically get cert reload. Consumers that previously wired `CertReloader` manually (e.g., Fleet Server in elastic/fleet-server#6838) can remove that manual wiring.

## Why is it important?

In environments with frequent certificate rotation — Kubernetes (where Secrets are remounted into pods), serverless (where cert-manager manages per-project certificates), and general ops — processes currently require a restart to pick up rotated TLS certificates. With this change, any component using `tlscommon` reloads certificates automatically, both as a TLS server and as a TLS client.

## Checklist

## Related issues

- Follow-up to #404 and #405
- Enables simplification of elastic/fleet-server#6838

### DIFF
--- a/transport/tlscommon/cert_reloader.go
+++ b/transport/tlscommon/cert_reloader.go
@@ -156,4 +156,3 @@ func (r *CertReloader) getCertificate() (*tls.Certificate, error) {
 	r.cert = &cert
 	return r.cert, nil
 }
-
--- a/transport/tlscommon/cert_reloader.go
+++ b/transport/tlscommon/cert_reloader.go
@@ -156,15 +156,9 @@ func (r *CertReloader) getCertificate() (*tls.Certificate, error) {
 	return r.cert, nil
 }
 
-// newCertReloaderFromConfig creates a CertReloader from a CertificateConfig
-// and CertificateReload configuration. It resolves the passphrase from the
-// config and applies the reload interval.
-func newCertReloaderFromConfig(certCfg CertificateConfig, reloadCfg CertificateReload) (*CertReloader, error) {
-	var opts []CertReloaderOption
-	if reloadCfg.ReloadInterval > 0 {
-		opts = append(opts, WithReloadInterval(reloadCfg.ReloadInterval))
-	}
-
+// newCertReloaderFromConfig creates a CertReloader from a CertificateConfig.
+// Additional options (e.g. WithReloadInterval) can be passed by the caller.
+func newCertReloaderFromConfig(certCfg CertificateConfig, opts ...CertReloaderOption) (*CertReloader, error) {
 	passphrase, err := certCfg.resolvePassphrase()
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve TLS key passphrase: %w", err)
--- a/transport/tlscommon/cert_reloader.go
+++ b/transport/tlscommon/cert_reloader.go
@@ -37,11 +37,12 @@ const defaultReloadInterval = 5 * time.Second
 // This design follows the OpenTelemetry Collector's configtls approach: no file
 // watchers or extra goroutines — just a time check on the handshake hot path.
 type CertReloader struct {
-	certPath       string
-	keyPath        string
-	passphrase     string
-	reloadInterval time.Duration
-	log            *logp.Logger
+	certPath         string
+	keyPath          string
+	passphrase       string
+	disableLegacyPEM bool
+	reloadInterval   time.Duration
+	log              *logp.Logger
 
 	mu         sync.RWMutex
 	cert       *tls.Certificate
@@ -63,6 +64,14 @@ func WithReloadInterval(d time.Duration) CertReloaderOption {
 func WithPassphrase(passphrase string) CertReloaderOption {
 	return func(r *CertReloader) {
 		r.passphrase = passphrase
+	}
+}
+
+// WithDisableLegacyPEMSupport configures whether encrypted PKCS#1 PEM keys
+// should be treated as an error instead of a deprecation warning.
+func WithDisableLegacyPEMSupport(disable bool) CertReloaderOption {
+	return func(r *CertReloader) {
+		r.disableLegacyPEM = disable
 	}
 }
 
@@ -94,12 +103,13 @@ func NewCertReloader(certPath, keyPath string, opts ...CertReloaderOption) (*Cer
 	return r, nil
 }
 
+// loadKeyPair mirrors the static cert-loading path in LoadCertificate (tls.go).
 func (r *CertReloader) loadKeyPair() (tls.Certificate, error) {
-	certPEM, err := ReadPEMFile(r.log, r.certPath, r.passphrase)
+	certPEM, err := readPEMFile(r.log, r.certPath, r.passphrase, r.disableLegacyPEM)
 	if err != nil {
 		return tls.Certificate{}, fmt.Errorf("reading certificate file: %w", err)
 	}
-	keyPEM, err := ReadPEMFile(r.log, r.keyPath, r.passphrase)
+	keyPEM, err := readPEMFile(r.log, r.keyPath, r.passphrase, r.disableLegacyPEM)
 	if err != nil {
 		return tls.Certificate{}, fmt.Errorf("reading key file: %w", err)
 	}
@@ -110,6 +120,17 @@ func (r *CertReloader) loadKeyPair() (tls.Certificate, error) {
 // reload interval has elapsed. It is safe for concurrent use and is intended
 // to be used with tls.Config.GetCertificate.
 func (r *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	return r.getCertificate()
+}
+
+// GetClientCertificate returns the current certificate, reloading from disk if
+// the reload interval has elapsed. It is safe for concurrent use and is
+// intended to be used with tls.Config.GetClientCertificate.
+func (r *CertReloader) GetClientCertificate(_ *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+	return r.getCertificate()
+}
+
+func (r *CertReloader) getCertificate() (*tls.Certificate, error) {
 	r.mu.RLock()
 	if time.Now().Before(r.nextReload) {
 		defer r.mu.RUnlock()
@@ -133,4 +154,27 @@ func (r *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate,
 	}
 	r.cert = &cert
 	return r.cert, nil
+}
+
+// newCertReloaderFromConfig creates a CertReloader from a CertificateConfig
+// and CertificateReload configuration. It resolves the passphrase from the
+// config and applies the reload interval.
+func newCertReloaderFromConfig(certCfg CertificateConfig, reloadCfg CertificateReload) (*CertReloader, error) {
+	var opts []CertReloaderOption
+	if reloadCfg.ReloadInterval > 0 {
+		opts = append(opts, WithReloadInterval(reloadCfg.ReloadInterval))
+	}
+
+	passphrase, err := certCfg.resolvePassphrase()
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve TLS key passphrase: %w", err)
+	}
+	if passphrase != "" {
+		opts = append(opts, WithPassphrase(passphrase))
+	}
+	if certCfg.DisableLegacyPEMSupport {
+		opts = append(opts, WithDisableLegacyPEMSupport(true))
+	}
+
+	return NewCertReloader(certCfg.Certificate, certCfg.Key, opts...)
 }
--- a/transport/tlscommon/cert_reloader.go
+++ b/transport/tlscommon/cert_reloader.go
@@ -31,7 +31,7 @@ const defaultReloadInterval = 5 * time.Second
 // CertReloader periodically reloads TLS certificate and key files from disk.
 // On each call to GetCertificate (i.e., on each TLS handshake), it checks
 // whether the reload interval has elapsed and, if so, re-reads the files from
-// disk. Invalid cert/key pairs are silently skipped, preserving the last
+// disk. Invalid cert/key pairs are logged and skipped, preserving the last
 // successfully loaded certificate.
 //
 // This design follows the OpenTelemetry Collector's configtls approach: no file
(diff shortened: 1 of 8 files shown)


cmacknz:
Enabling this by default means if we backport this elastic-agent-libs version we'll be introducing this in patch releases.

I think we'll want this on by default from the next minor onwards, and disabled in each existing release branch.

transport/tlscommon/cert_reloader.go
```diff
@@ -134,3 +155,26 @@ func (r *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate,
...
 }
+
+// newCertReloaderFromConfig creates a CertReloader from a CertificateConfig
+// and CertificateReload configuration. It resolves the passphrase from the
+// config and applies the reload interval.
+func newCertReloaderFromConfig(certCfg CertificateConfig, reloadCfg CertificateReload) (*CertReloader, error) {
+	var opts []CertReloaderOption
+	if reloadCfg.ReloadInterval > 0 {
+		opts = append(opts, WithReloadInterval(reloadCfg.ReloadInterval))
+	}
```
AndersonQ: Why doesn't it take in a `CertReloaderOption` directly?

transport/tlscommon/tls_config.go
```diff
@@ -143,6 +148,14 @@ func (c *TLSConfig) ToConfig() *tls.Config {
 		Time:               c.time,
 		VerifyConnection:   makeVerifyConnection(c, c.Logger),
 	}
+
+	if c.certReloader != nil {
+		cfg.GetCertificate = c.certReloader.GetCertificate
+		cfg.GetClientCertificate = c.certReloader.GetClientCertificate
+		cfg.Certificates = nil
```
AndersonQ: [Question]
With cretReload enabled by default, all callers will get a nil `cfg.Certificates`. Can't it cause a problem for the users? Code that does not account for cretReload now does not have any certificate when calling `ToConfig`.
Unless we're 100% sure it won't happen, perhaps the best here would keep the certificate to be backwards compatible.

Or am I missing something?

mauri870 (dismissed):

transport/tlscommon/tls_config.go
blakerouse: I think it would be best not to clear `cfg.Certificates` if that is possible. I believe that is not done here though, because goland TLS will use `Certificates` over `GetCertificate`. That means that code that uses this library will need to check if `GetCertificate == nil` that it doesn't set `Certificates`.

I don't know if that is really better. Where its probably just better to allow them to assign all three and move on. But that does meant that all projects that use this library, need to know about this change when they bump the version.

ycombinator: Good discussion, guys. I've removed this line in 6b0418c because in the cert reloading path, `Certificates` is already set to `nil` (zero value of slice) during construction: 
- [config.go](https://github.com/ycombinator/elastic-agent-libs/blob/bedf6d5b8f9dd2ee12d8c7286e56e7cbb2066724/transport/tlscommon/config.go#L68-L81)
- [server_config.go](https://github.com/ycombinator/elastic-agent-libs/blob/bedf6d5b8f9dd2ee12d8c7286e56e7cbb2066724/transport/tlscommon/server_config.go#L96-L109)

When cert reloading is not enabled, we don't enter this `if` block and `Certificates` is set to `c.Certificates` above this `if` block.  So no change in behavior there from prior to this PR.

---

Now, the question is: are there any callers of `ToConfig()` that _don't_ simply pass along the returned `cfg` to Go's TLS mechanism but, instead, use `Certificates` directly.  I had Claude check call sites in elastic-agent-libs, elastic-agent, beats, and fleet-server and nowhere is `Certificates` being read from:

### `elastic-agent-libs` (3 call sites, all pass-through)
- [`tls_config.go:193`](https://github.com/ycombinator/elastic-agent-libs/blob/cert-reload-wiring/transport/tlscommon/tls_config.go#L193) — `cc.ToConfig()` in
   `BuildModuleClientConfig`
- [`tls_config.go:216`](https://github.com/ycombinator/elastic-agent-libs/blob/cert-reload-wiring/transport/tlscommon/tls_config.go#L216) — `c.ToConfig()` in
  `BuildServerConfig`
- [`httpcommon.go:311`](https://github.com/ycombinator/elastic-agent-libs/blob/cert-reload-wiring/transport/httpcommon/httpcommon.go#L311) — `tls.ToConfig()`
  assigned to `t.TLSClientConfig`

### `elastic-agent` (1 call site)
- [`internal/pkg/otel/translate/common.go:216`](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/translate/common.go#L216) — reads `CipherSuites`, `CurvePreferences`, `MinVersion`, `MaxVersion` only

### `fleet-server` (2 call sites)
- [`internal/pkg/config/output.go:130`](https://github.com/elastic/fleet-server/blob/main/internal/pkg/config/output.go#L130) — `tls.ToConfig()` assigned to `httpTransport.TLSClientConfig` (pass-through)
- [`internal/pkg/api/server.go:144`](https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/server.go#L144) — *writes* `srv.TLSConfig.Certificates = nil` (clears it, does not read)

### `beats` (15 call sites across 13 files)
- [`filebeat/input/redis/input.go:67`](https://github.com/elastic/beats/blob/main/filebeat/input/redis/input.go#L67) — pass-through
- [`metricbeat/module/mysql/mysql.go:72`](https://github.com/elastic/beats/blob/main/metricbeat/module/mysql/mysql.go#L72) — pass-through
- [`metricbeat/module/aerospike/aerospike.go:58`](https://github.com/elastic/beats/blob/main/metricbeat/module/aerospike/aerospike.go#L58) — pass-through
- [`x-pack/libbeat/common/aws/credentials.go:97`](https://github.com/elastic/beats/blob/main/x-pack/libbeat/common/aws/credentials.go#L97) — pass-through
- [`x-pack/filebeat/input/streaming/websocket.go:594`](https://github.com/elastic/beats/blob/main/x-pack/filebeat/input/streaming/websocket.go#L594) — pass-through
- [`libbeat/processors/translate_ldap_attribute/translate_ldap_attribute.go:110`](https://github.com/elastic/beats/blob/main/libbeat/processors/translate_ldap_attribute/translate_ldap_attribute.go#L110) — pass-through
- [`metricbeat/module/mongodb/mongodb.go:162`](https://github.com/elastic/beats/blob/main/metricbeat/module/mongodb/mongodb.go#L162) — pass-through
- [`metricbeat/module/redis/metricset.go:58`](https://github.com/elastic/beats/blob/main/metricbeat/module/redis/metricset.go#L58) — pass-through
- [`metricbeat/module/mysql/query/query.go:90`](https://github.com/elastic/beats/blob/main/metricbeat/module/mysql/query/query.go#L90) — pass-through
- [`x-pack/osquerybeat/internal/install/artifact/artifact.go:229`](https://github.com/elastic/beats/blob/main/x-pack/osquerybeat/internal/install/artifact/artifact.go#L229) — pass-through
- [`libbeat/processors/add_cloud_metadata/providers.go:172`](https://github.com/elastic/beats/blob/main/libbeat/processors/add_cloud_metadata/providers.go#L172) — pass-through
- [`x-pack/libbeat/common/cloudfoundry/hub.go:139`](https://github.com/elastic/beats/blob/main/x-pack/libbeat/common/cloudfoundry/hub.go#L139) — pass-through
- [`x-pack/libbeat/common/cloudfoundry/hub.go:163`](https://github.com/elastic/beats/blob/main/x-pack/libbeat/common/cloudfoundry/hub.go#L163) — reads `InsecureSkipVerify` only
- [`x-pack/otel/oteltranslate/tls_otel.go:127`](https://github.com/elastic/beats/blob/main/x-pack/otel/oteltranslate/tls_otel.go#L127) — reads `CipherSuites`, `MinVersion`, `MaxVersion` only

**Zero call sites across all four repos read `.Certificates` from the returned `*tls.Config`.**

transport/tlscommon/types.go
```diff
@@ -389,6 +390,22 @@ type CertificateConfig struct {
...
+	}
+	if c.PassphrasePath != "" {
+		p, err := os.ReadFile(c.PassphrasePath)
+		if err != nil {
+			return "", fmt.Errorf("unable to read key passphrase file: %w", err)
+		}
+		return string(p), nil
+	}
+	return "", nil
+}
```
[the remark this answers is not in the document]
ycombinator: Fixed in a7be6e8

transport/tlscommon/cert_reloader.go
ycombinator: Fixed in 25cca984c905bf65f681750159ce19b3302c3a39.

transport/tlscommon/cert_reloader.go
```diff
@@ -134,3 +155,20 @@ func (r *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate,
 	r.cert = &cert
 	return r.cert, nil
 }
+
+// newCertReloaderFromConfig creates a CertReloader from a CertificateConfig.
+// Additional options (e.g. WithReloadInterval) can be passed by the caller.
+func newCertReloaderFromConfig(certCfg CertificateConfig, opts ...CertReloaderOption) (*CertReloader, error) {
```
AndersonQ: [Suggestion]

I won't block for that, but now the API allows 2 different ways to configure the same thing. The caller could pass `WithPassphrase(passphrase), WithDisableLegacyPEMSupport(true)` and it's not clear which one has precedence. 
I get this function is sugar syntax to avoid duplication. 

Strictly speaking, I believe it should either take only `opts ...CertReloaderOption` or the params for passphrase, DisableLegacyPEMSupport and CertificateReload.ReloadInterval.

this is a private function, so less chance of misuse, as I said, I won't block for it.

transport/tlscommon/cert_reloader.go
AndersonQ, line 161: I know it isn't your change, there is a logger here, shouldn't it log there was an issue here?
Now with enabled by default, if anything happens, the cers aren't reloaded, it uses the old ones, they become invalid, how would one know what is happening, why the old certs are being used?

AndersonQ (commented):
There is a missmatch on how the passphrase is read:
- [resolvePassphrase](https://github.com/elastic/elastic-agent-libs/pull/417/changes#diff-94f3ec87c2a4d21678769021c19fcbc273b1acc15dcc68e24b45b57229ee3838R405) trims it
- [LoadCertificate](https://github.com/ycombinator/elastic-agent-libs/blob/cert-reload-wiring/transport/tlscommon/tls.go#L50-56) does not

I had a quick look, and it seems the standard, [what openssl does](https://docs.openssl.org/master/man1/openssl-passphrase-options/), is aligend with timming, however it makes both function differ in behaviour, that is the real issue here. 
                   
As I believe letting a line break there is a mistake and this change is already                     
changing default behaviour, I think you could align `LoadCertificate` and make it                   
also trim what it reads. Even better, make a function to read the passphrase and                    
both `resolvePassphrase` and `LoadCertificate` use it.
What do you think?

ycombinator:
> Even better, make a function to read the passphrase and both `resolvePassphrase` and `LoadCertificate` use it.

Yep; done in 9913b4d.

transport/tlscommon/cert_reloader.go
ycombinator: Good idea.  Done in a2e6a99.

transport/tlscommon/cert_reloader.go
ycombinator: Good catch. Fixed to only use `opts` in 3c81579.

transport/tlscommon/cert_reloader.go
```diff
@@ -129,6 +150,7 @@ func (r *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate,
 
 	cert, err := r.loadKeyPair()
 	if err != nil {
+		r.log.Errorf("Failed to reload TLS certificate from %s and %s, continuing with current certificate: %v", r.certPath, r.keyPath, err)
```
AndersonQ, line 159: [Blocker]
Sorry, I asked for a log, but now I realised there is an serious issue on the `CertReloader`. It accepts either a path or a key/cert, and both work. 
The flexibility of the lib is a troublemaker :/

If a PEM is passed instead of a path, it just works, the reload does not reload anything and here it could log the plain key.

I'm wondering, I don't think accepting a PEM instead of the path is the desired behaviour, so a validation on `NewCertReloader` should fix that. It can use `IsPEMString` just as `NewPEMReader` does and error if either is a PEM.

I made a test that shows that: 

[test.patch](https://github.com/user-attachments/files/27515113/test.patch)

mauri870 (commented):
LGTM! I'm using this as part of the work in cert reloading for the beatsauthextension component, haven't found any issues while using your fork! Lets address Anderson's comments and get his approval before merging.

mauri870 (dismissed):
LGTM! I'm using this as part of the work in cert reloading for the beatsauthextension component, haven't found any issues while using your fork! Lets address Anderson's comments and get his approval before merging.

transport/tlscommon/cert_reloader.go
ycombinator: Great catch, @AndersonQ! Fixed in 99a833f.

mauri870 (approved):

PR #424, opened 2026-05-22
[config] Implement confmap.Unmarshal interface

VihasMakwana:
## What does this PR do?

Beats deal with `config.C`, whereas OTeL equivalent is `confmap.Conf`. As of now, we convert confmap to a string map, and create a new `config.C` for our use case.
This PR implements the `confmap.Unmarshal` interface, allowing the collector to populate `config.C` directly while reading the configuration.


config/config.go
```diff
@@ -27,12 +27,15 @@ import (
 	"github.com/elastic/elastic-agent-libs/str"
 	ucfg "github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/yaml"
+	"go.opentelemetry.io/collector/confmap"
```
khushijain21, line 30: so every project that uses this package will now have otel dependency in its module graph. We should avoid this

VihasMakwana:
Collector core might need some additional changes for this to work correctly. I'll move to draft for now.

PR #425, opened 2026-06-18
fix(httpcommon): preserve TLS metadata for HTTPS via custom dialer

shahzad31:
## TL;DR

`RoundTripper` installs a custom `DialTLSContext`, but it wrapped the resulting `*tls.Conn` in a logging/stats connection. `net/http` only populates `Response.TLS` when the connection from `DialTLSContext` is concretely a `*tls.Conn`, so TLS metadata was silently dropped for **every** HTTPS request through this transport. The fix moves the wrappers under the TLS layer so the dialer returns a real `*tls.Conn` again.

## What does this PR do?

`HTTPTransportSettings.RoundTripper` builds an `*http.Transport` and sets a custom `DialTLSContext` (`tlsDialer.DialContext`). The dialer wrappers were applied in the wrong order:

```
Before:  net/http  ←  loggingConn  ←  (statsConn)  ←  *tls.Conn  ←  TCP socket
                       └── outermost conn is NOT a *tls.Conn ──┘

After:   net/http  ←  *tls.Conn  ←  loggingConn  ←  (statsConn)  ←  TCP socket
                       └── outermost conn IS a *tls.Conn ──┘
```

- **Before:** logging (always on, since a default logger is set when none is provided) and the optional `WithIOStats` dialer wrapped the `*tls.Conn`, so `net/http` received a `*loggingConn` / `*statsConn`.
- **After:** those wrappers are applied to the **plain TCP dialer**, then TLS wraps that. The TLS dialer stays the outermost layer and returns a real `*tls.Conn`, while logging/stats keep operating on the underlying socket.

### Why `Response.TLS` was `nil`

`net/http` records the TLS connection state — and therefore sets `Response.TLS` — only when the connection returned by a custom `DialTLSContext` is concretely a `*tls.Conn`:

```go
// (*http.Transport).dialConn
pconn.conn, err = t.customDialTLS(ctx, "tcp", cm.addr())
if tc, ok := pconn.conn.(*tls.Conn); ok { // a wrapped conn fails this assertion
    ...
    pconn.tlsState = &cs               // → later assigned to resp.TLS
}
```

A wrapped connection fails the `*tls.Conn` assertion, so `pconn.tlsState` (and `Response.TLS`) stayed `nil`.

## Why is it important?

Any consumer relying on `Response.TLS` loses all TLS metadata for HTTPS made through a custom dialer. Concretely, **Heartbeat HTTP monitors with `max_redirects > 0` (or a proxy) stopped emitting `tls.*` fields** after the 8.0 migration to `elastic-agent-libs`, breaking TLS certificate-expiry monitoring for redirecting endpoints (elastic/beats#48335). This restores the pre-8.0 behaviour.

## Behaviour change (release note)

For HTTPS connections, `WithIOStats` byte counts now reflect on-the-wire encrypted traffic (including the TLS handshake) instead of decrypted application-layer bytes, because the stats dialer now sits beneath the TLS layer.

## Testing

Added `Test_HTTPTransportSettings_RoundTripper_TLSMetadata`:

- Passes with the fix; fails on `main` (reproduces the regression).
- Covers both the default (logging-only) path and the `WithIOStats` path.
- Verifies I/O stats are still recorded over TLS (now counted at the socket layer rather than the plaintext layer).

```
go test ./transport/...   # passes
```

## Checklist

## Related issues

- Relates elastic/beats#48335
- Regression test (beats): elastic/beats#51339

### DIFF
--- a/transport/httpcommon/httpcommon.go
+++ b/transport/httpcommon/httpcommon.go
@@ -265,10 +265,6 @@ func (settings *HTTPTransportSettings) RoundTripper(opts ...TransportOption) (ht
 		return nil, err
 	}
 
-	// Apply stats and logging to the TCP dialer before wrapping it with TLS, so
-	// the TLS dialer stays outermost and returns a *tls.Conn. net/http only sets
-	// Response.TLS when DialTLSContext yields a *tls.Conn (it does conn.(*tls.Conn)
-	// in Transport.dialConn); wrapping it would drop all tls.* metadata.
 	for _, opt := range opts {
 		if dialOpt, ok := opt.(dialerModOption); ok {
 			dialer = dialOpt.applyDialer(settings, dialer)
@@ -279,6 +275,10 @@ func (settings *HTTPTransportSettings) RoundTripper(opts ...TransportOption) (ht
 		dialer = transport.LoggingDialer(dialer, logger)
 	}
 
+	// Wrap the TCP dialer (already decorated with stats/logging above) with TLS
+	// last, so the TLS dialer stays outermost and returns a *tls.Conn. net/http
+	// only sets Response.TLS when DialTLSContext yields a *tls.Conn (it does
+	// conn.(*tls.Conn) in Transport.dialConn); wrapping it would drop tls.* metadata.
 	tlsDialer := transport.TLSDialer(dialer, tls, settings.Timeout, extra.logger)
 
 	var rt http.RoundTripper
--- a/transport/httpcommon/httpcommon.go
+++ b/transport/httpcommon/httpcommon.go
@@ -265,18 +265,10 @@ func (settings *HTTPTransportSettings) RoundTripper(opts ...TransportOption) (ht
 		return nil, err
 	}
 
-	// Apply dialer modifications (e.g. I/O stats) and connection logging to the
-	// plain TCP dialer *before* wrapping it with TLS, so the TLS dialer remains
-	// the outermost layer and returns a *tls.Conn.
-	//
-	// net/http only records the TLS connection state (and therefore populates
-	// Response.TLS) when the connection produced by Transport.DialTLSContext is a
-	// *tls.Conn (see (*http.Transport).dialConn, which does a concrete-type
-	// assertion `conn.(*tls.Conn)`). Wrapping the *tls.Conn in any other net.Conn
-	// implementation (stats or logging) makes that assertion fail and silently
-	// drops all tls.* metadata for HTTPS requests. Wrapping the underlying TCP
-	// connection instead keeps both behaviours: connection-level logging/stats and
-	// a usable *tls.Conn for net/http.
+	// Apply stats and logging to the TCP dialer before wrapping it with TLS, so
+	// the TLS dialer stays outermost and returns a *tls.Conn. net/http only sets
+	// Response.TLS when DialTLSContext yields a *tls.Conn (it does conn.(*tls.Conn)
+	// in Transport.dialConn); wrapping it would drop all tls.* metadata.
 	for _, opt := range opts {
 		if dialOpt, ok := opt.(dialerModOption); ok {
 			dialer = dialOpt.applyDialer(settings, dialer)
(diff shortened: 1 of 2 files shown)


transport/httpcommon/httpcommon_test.go
```diff
@@ -336,6 +339,76 @@ func Test_HTTPTransportSettings_RoundTripper(t *testing.T) {
...
+	srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	ca := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: srv.TLS.Certificates[0].Certificate[0],
+	})
+	require.NotEmpty(t, ca)
```
mcapell: nit: I don't think `ca` can be empty here, so there is no need to do this check.

transport/httpcommon/httpcommon_test.go
```diff
@@ -336,6 +339,76 @@ func Test_HTTPTransportSettings_RoundTripper(t *testing.T) {
...
+	})
+
+	t.Run("populates Response.TLS and records stats with WithIOStats", func(t *testing.T) {
+		stats := &countingStatser{}
+		rt, err := newSettings().RoundTripper(WithIOStats(stats))
+		require.NoError(t, err)
+
+		tlsState := doRequest(t, rt)
+		require.NotNil(t, tlsState, "Response.TLS must be set for HTTPS requests")
+		require.NotEmpty(t, tlsState.PeerCertificates, "TLS peer certificates must be present")
```
mcapell: nit / non-blocking:

this is already checked in the previous test. We could merge both subtests into one

transport/httpcommon/httpcommon.go
```diff
@@ -265,19 +265,22 @@ func (settings *HTTPTransportSettings) RoundTripper(opts ...TransportOption) (ht
 		return nil, err
 	}
 
-	tlsDialer := transport.TLSDialer(dialer, tls, settings.Timeout, extra.logger)
+	// Apply stats and logging to the TCP dialer before wrapping it with TLS, so
```
mcapell: nit: the comment without the diff is a bit confusing. I'd move the comment to the `tlsDialer := transport.TLSDialer(...)`

mcapell (approved):
LGTM.

Added a few non-blocking comments

transport/httpcommon/httpcommon_test.go
shahzad31: Good catch, removed in 8de2329 — the PEM-encoded test cert can't be empty.

transport/httpcommon/httpcommon_test.go
shahzad31: Merged in 8de2329. The `WithIOStats` path already exercises the always-on logging dialer, so a single request now covers both the TLS metadata and the I/O stats assertions.

transport/httpcommon/httpcommon.go
shahzad31: Moved next to the `tlsDialer := transport.TLSDialer(...)` assignment in 8de2329.

belimawr (approved):
LGTM. I just have one question regarding the stats now. Because the TLSDialer now wraps the `StatsDialer`, the metrics returned now include the TLS handshake. I cannot tell whether this is an issue for any user of `eastic-agent-libs`, but it would be good to at least mention it in the commit message and "release notes".

shahzad31:
Good point @belimawr. You're right — with the stats dialer now beneath TLS, `WithIOStats` counts on-the-wire encrypted bytes (including the handshake) for HTTPS, whereas before it saw decrypted application-layer bytes. This is arguably the more accurate measure of actual network usage, but it is a behaviour change for stats consumers.

I've added a concise "Behaviour change (release note)" section to the PR description. Since the squash-merge commit message is taken from the PR title + body, that note will land in the commit message too.

AndersonQ (approved):

PR #426, opened 2026-06-22
feat(tlscommon): enforce FIPS 140-3 peer cert key types across all TLS verification modes

macdewee:
## What does this PR do?

Part of the transition from Microsoft Go to native Go FIPS 140-3.

Native Go skips its FIPS certificate key-type check when a custom TLS callback is installed ([golang/go#80074](https://github.com/golang/go/issues/80074)). This codebase uses custom callbacks everywhere, so the check was silently bypassed. This PR adds the check back explicitly.

Enforcement targets the full certificate chain — not just the leaf — because every link in the chain is part of the connection's security contract. If any certificate in any valid path uses a non-approved key, the connection is rejected.

Both client and server verification paths are covered.

## Why is it important?

Microsoft Go enforces FIPS at the crypto library level, so this gap didn't exist before. Switching to native Go exposed it: peers could present certificates with non-FIPS key types (e.g. RSA-1024) and the handshake would succeed in a FIPS build.

Non-FIPS builds are not affected — all changes are behind the `requirefips` build tag.

## Checklist

## Author's Checklist

- [x] All verification modes covered on both client and server side
- [x] Full chain checked, not just the leaf certificate
- [x] Non-FIPS builds verified unaffected
- [x] Known limitation documented: `RequireAnyClientCert` mode — Go skips chain building for it, so only the presented certificates are checked

## Related issues

- Relates https://github.com/golang/go/issues/80074


macdewee:
Switching to draft, as there is ongoing discussion in the upstream. I am learning new nuances and the final shape may look slightly different

edit: ready to review

transport/tlscommon/tls_config.go
```diff
@@ -317,17 +317,18 @@ func makeVerifyConnection(cfg *TLSConfig, logger *logp.Logger) func(tls.Connecti
...
 	switch cfg.Verification {
+	case VerifyNone:
+		// Chain and hostname verification are disabled, but FIPS key-type constraints
+		// still apply — all peer certificates are checked for approved key types.
+		return fipsVerifyNoneCallback()
 	case VerifyFull:
-		// Cert is trusted by CA
-		// Hostname or IP matches the certificate
-		// tls.Config.InsecureSkipVerify  is set to true
+		// Cert is trusted by CA; hostname or IP matches the certificate.
```
belimawr: [Question]
Is there any reason to remove those comments? Given this is such a critical piece of software requiring specialised knowledge, I believe the more detailed explanation is helpful.

transport/tlscommon/tls_config.go
```diff
@@ -317,17 +317,18 @@ func makeVerifyConnection(cfg *TLSConfig, logger *logp.Logger) func(tls.Connecti
...
 		return func(cs tls.ConnectionState) error {
 			if cfg.CATrustedFingerprint != "" {
 				if err := trustRootCA(cfg, cs.PeerCertificates, logger); err != nil {
 					return err
 				}
 			}
-			// On the client side, PeerCertificates can't be empty.
```
belimawr, line 330: Same thing here, this could have been kept

transport/tlscommon/tls_config.go
belimawr: The same applies to the rest of the file.

transport/tlscommon/tls_config.go
```diff
@@ -386,94 +389,151 @@ func makeVerifyConnection(cfg *TLSConfig, logger *logp.Logger) func(tls.Connecti
...
-				return verifyCAPin(cfg.CASha256, cs.VerifiedChains)
+		// Static CAs: the TLS stack builds a verified chain for RequireAndVerifyClientCert
+		// and VerifyClientCertIfGiven; FIPS checks use those chains.
+		// For RequireAnyClientCert, chain building is deliberately skipped, so only
+		// the certificates the client chose to send are checked. A client that omits
+		// intermediates can pass with a non-FIPS intermediate — accepted limitation;
+		// use RequireAndVerifyClientCert for full chain coverage.
+		return func(cs tls.ConnectionState) error {
+			if len(cs.PeerCertificates) == 0 {
+				return noPeerCertsError(cfg)
```
belimawr, line 469: This check is here and in the `len(cfg.CASha256) > 0` branch. Can we move it before the if statement so we don't duplicate the check?

transport/tlscommon/tls_fips_test.go
```diff
@@ -20,18 +20,543 @@
...
+					tlsConn := tls.Server(conn, serverNativeCfg)
+					if err := tlsConn.Handshake(); err == nil {
+						// Write a sentinel byte so the client can distinguish
+						// acceptance from a TLS 1.3 rejection that only arrives
+						// on the first application-data read.
+						_, _ = tlsConn.Write([]byte{0})
+					}
+					tlsConn.Close()
+				}
+			}()
```
belimawr: Nit, you can use the new (Go 1.25) `WaitGroup.Go`
```suggestion
			wg.Go(func() {
				for {
					conn, err := ln.Accept()
					if err != nil {
						return
					}
					tlsConn := tls.Server(conn, serverNativeCfg)
					if err := tlsConn.Handshake(); err == nil {
						// Write a sentinel byte so the client can distinguish
						// acceptance from a TLS 1.3 rejection that only arrives
						// on the first application-data read.
						_, _ = tlsConn.Write([]byte{0})
					}
					tlsConn.Close()
				}
			})
```

transport/tlscommon/tls_config.go
macdewee: Makes sense, I restored the original comments

ycombinator (approved):

belimawr (approved):

PR #429, opened 2026-07-01
Add opt-in header-value expansion to the redact package

straistaru:
## What does this PR do?

Adds `WithHeaderValueKeys(...)` so specific map keys can hold comma-separated HTTP headers (`Name=Value,Name=Value`).

During redaction, those values are parsed, each header name is checked with the existing redactKey rules, and the original key is replaced with expanded entries like `FLEET_HEADER::X-Elastic-App-Auth`. Sensitive headers become `REDACTED`; others keep their values.

## Why is it important?

Environment variables can store HTTP headers in a single string field (e.g. `FLEET_HEADER`). The redactor only looked at map keys before, so secrets inside those strings could leak into diagnostics. This lets us redact sensitive headers without wiping the whole field.

Example of problematic ENV var:
```yaml
FLEET_HEADER: X-Elastic-App-Auth=eyJhbGciOiJSUzI1NiJ9... 
```

Result after redaction:
```yaml
FLEET_HEADER: REDACTED # to understand that this is the original value, and the variable was set
FLEET_HEADER::X-Elastic-App-Auth: REDACTED # to understand each individual header that was set
```

Please check the added test cases to get an overview of how this applies in multiple scenarios.

## Checklist

## Author's Checklist

- [ ]

## Related issues

-

### DIFF
--- a/redact/redact.go
+++ b/redact/redact.go
@@ -40,7 +40,7 @@ type redactOptions struct {
 	errOut          io.Writer
 	markerPrefix    string
 	ignoreKeys      []string
-	headerValueKeys []string
+	headerValueKeys map[string]struct{}
 }
 
 type headerPair struct {
@@ -49,8 +49,9 @@ type headerPair struct {
 }
 
 type headerExpansion struct {
-	origKey string
-	pairs   []headerPair
+	origKey     string
+	pairs       []headerPair
+	anyRedacted bool
 }
 
 // WithErrorOutput determines where any error messages that are encountered are written to.
@@ -84,7 +85,12 @@ func WithIgnoreKeys(keys ...string) RedactOption {
 // replaced with REDACTED.
 func WithHeaderValueKeys(keys ...string) RedactOption {
 	return func(ro *redactOptions) {
-		ro.headerValueKeys = append(ro.headerValueKeys, keys...)
+		if ro.headerValueKeys == nil {
+			ro.headerValueKeys = make(map[string]struct{}, len(keys))
+		}
+		for _, k := range keys {
+			ro.headerValueKeys[k] = struct{}{}
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


redact/redact.go
```diff
@@ -149,6 +177,17 @@ func redactMap[K comparable](obj map[K]any, ro *redactOptions) {
 					val = REDACTED
 				} else if redactedValue, redact := redactURL(cast); redact {
 					val = redactedValue
+				} else if len(ro.headerValueKeys) > 0 {
```
michalpristas: nit: we can do the check upfront and use computed value to save some cycles instead of on each env var

redact/redact.go
```diff
@@ -149,6 +177,17 @@ func redactMap[K comparable](obj map[K]any, ro *redactOptions) {
...
+					if keyString, ok := any(key).(string); ok && slices.Contains(ro.headerValueKeys, keyString) {
+						pairs := parseHeaderPairs(cast)
+						for i := range pairs {
+							if redactKey(pairs[i].name, ro) {
+								pairs[i].value = REDACTED
+							}
```
michalpristas, line 196: what about redacting header value itself if we redact something. i see you're removing original header value. i'm ok with this, it's just keeping the value tells us there was in fact a header env var. 
can we add test verifying either of the behavior. whatever you settle on

michalpristas (dismissed):

redact/redact.go
straistaru: > what about redacting header value itself if we redact something. i see you're removing original header value. i'm ok with this, it's just keeping the value tells us there was in fact a header env var.

Do you mean that env vars like:
```yaml
FLEET_HEADER: X-Elastic-App-Auth=jwt1
```
Will produce both?:
```yaml
FLEET_HEADER: REDACTED
FLEET_HEADER::X-Elastic-App-Auth: REDACTED
```

michalpristas: yes

redact/redact.go
straistaru: Fixed in https://github.com/elastic/elastic-agent-libs/pull/429/commits/beb4efa85e7ecb20eb5c60344358f3f26b5c83f6

redact/redact.go
straistaru: Fixed in https://github.com/elastic/elastic-agent-libs/pull/429/commits/beb4efa85e7ecb20eb5c60344358f3f26b5c83f6

Please check all the test cases for a quick overview on how it will work.
TLDR: I kept the main env var as REDACTED besides the expanded values. Also I kept only the main env var with original value with no expansion if no redaction is necessary

michalpristas (approved):

khushijain21 (approved):
LGTM

PR #431, opened 2026-07-03
tlscommon: better logging certificate error

AndersonQ:
## What does this PR do? / Commit message

```
tlscommon: better logging certificate error                                                 
                                                                                                       
LoadCertificate and LoadCertificateAuthorities echoed the raw certificate/key                          
string into returned errors and debug logs when given a malformed PEM.                              
IsPEMString only checks for a leading "-", so a malformed inline key that lost                         
its leading dashes was misclassified as a file path and handed to os.Open,                             
whose *PathError echoes its argument verbatim — leaking key material into logs                      
and errors.                                                                                         
                                                                                                       
Add isInlinePEM, a stronger predicate that also treats multi-line strings and                          
PEM armor ("-----") as inline content, and pemSource, which passes a real path                         
through unchanged but redacts inline content to "PEM REDACTED". Apply pemSource                        
at every error/log site in LoadCertificate, LoadCertificateAuthorities, and                            
CertReloader, and switch NewPEMReader, NewCertReloader, and the reload-skip                            
guards in config.go/server_config.go from IsPEMString to isInlinePEM so malformed                      
inline content never reaches os.Open in the first place. Genuine missing file                          
paths are still reported verbatim so users can debug them.                                          
                                                                                                       
Add regression tests covering the leak on the cert, key, and CA paths,                                 
isInlinePEM/pemSource unit tests, and guards that valid inline PEM and real                            
missing-file paths still behave as before.                                                          
                                                                                                       
Assisted-by: Claude code                                                                               
```

## Checklist

## Related issues

- N/A


transport/tlscommon/tls.go
```diff
@@ -72,9 +72,10 @@ func LoadCertificate(config *CertificateConfig) (*tls.Certificate, error) {
 	// Do not log the key if it was provided as a string in the configuration to avoid
 	// leaking private keys in the debug logs. Log when the key is a file path.
 	if IsPEMString(key) {
-		log.Debugf("Loading certificate: %v with key from PEM string in config", certificate)
+		log.Debugf("Loading certificate: %v with key from PEM string in config", pemSource(certificate))
 	} else {
-		log.Debugf("Loading certificate: %v and key %v", certificate, key)
+		log.Debugf("Loading certificate: %v and key %v", certificate,
+			pemSource(key))
 	}
```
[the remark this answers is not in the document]
AndersonQ: but the certificate isn't privet, only the key. So is it indeed an issue?

transport/tlscommon/tls.go
```diff
@@ -262,3 +263,29 @@ func IsPEMString(s string) bool {
...
+// IsPEMString) so a path with a trailing newline -- e.g. from a YAML block
+// scalar -- is not misclassified as inline content. Used to keep key material
+// out of os.Open errors and log lines: a malformed inline key that lost its
+// leading dashes must not be mistaken for a path and handed to os.Open, which
+// would echo the whole blob.
+func isInlinePEM(s string) bool {
+	trimmed := strings.TrimSpace(s)
+	return IsPEMString(s) ||
+		strings.ContainsAny(trimmed, "\r\n") ||
+		strings.Contains(trimmed, "-----")
```
cmacknz: This is still going to miss things like copy paste errors where somebody copy pastes a PEM string but forgets one or more of the trailing dashes in the case where we cannot distinguish between an inline PEM and a possibly valid file path.

transport/tlscommon/cert_reloader_test.go
```diff
@@ -82,6 +82,37 @@ func TestNewCertReloader_EmptyPaths(t *testing.T) {
 	assert.Error(t, err)
 }
 
+// TestNewCertReloader_RejectsInlinePEM ensures inline PEM content — including a
+// malformed blob that lost its leading dashes — is rejected at construction, so
+// a private key can never be stored on the reloader's path fields and later
+// leaked to logs on a reload failure (ingest-dev#6152).
```
cmacknz:
```suggestion
// leaked to logs on a reload failure.
```

Let's keep private issues out of public source code.

transport/tlscommon/tls.go
```diff
@@ -53,14 +53,14 @@ func LoadCertificate(config *CertificateConfig) (*tls.Certificate, error) {
...
-		return nil, fmt.Errorf("%w %v", err, certificate)
+		log.Errorf("Failed reading certificate %s: %v", pemSource(certificate), err)
+		return nil, fmt.Errorf("%w (%s)", err, pemSource(certificate))
 	}
 
 	keyPEM, err := readPEMFile(log, key, passphrase, config.DisableLegacyPEMSupport)
 	if err != nil {
-		log.Errorf("Failed reading key file: %+v", err)
-		return nil, fmt.Errorf("%w %v", err, key)
+		log.Errorf("Failed reading key (%s): %v", pemSource(key), err)
```
cmacknz: Can we avoid logging `key` directly anywhere at all? `pemSource` still relies on a heuristic so having `pemSource(key)` isn't necessarily enough.

We still have the problem of not including the key in the error potentially returned by `os.Open`, since we didn't allow the user to provide whether what we are parsing is supposed a key or a file path. A slightly malformed PEM file also turns out to be a potentially valid file path as well, unfortunately: https://go.dev/play/p/wgQ3QwzDZYM

cmacknz: I think in the PEM Reader implementation at https://github.com/elastic/elastic-agent-libs/blob/a3e1e3b6287e0816695bab0e6d3773aa3485c6f8/transport/tlscommon/tls.go#L238-L241

You should check if the returned error is a Path error and drop the Path component from it. Something like the following, though I think it could be better to create a new `fs.PathError` to return with `Path` empty so that type assertions against `fs.PathError` higher up the chain would still work.

```go
r, err := os.Open(certificate)
if err != nil {
    var pathErr *fs.PathError
    if errors.As(err, &pathErr) {
        return nil, fmt.Errorf("%s: %w", pathErr.Op, pathErr.Err) // drop Path entirely
    }
    return nil, err
}
```

transport/tlscommon/tls.go
cmacknz: Talking with Claude about this, since PEM file content can be a valid path potentially, we could strengthen this heuristic to look for things that are highly unlikely to be user provided file paths:

> Strengthen isInlinePEM with a length + charset check, not just armor/newlines:
> strings.Trim(trimmed, base64Alphabet) == "" && len(trimmed) >= minSecretLen && !strings.ContainsAny(trimmed, "/\\.:")
> i.e., "long, pure base64 alphabet, no path-typical separators or extension" → treat as suspect and redact, same as armored content. The shortest realistic key encoding (an Ed25519 seed) is ~44 base64 chars; RSA/EC keys in practice run into the hundreds or low thousands. A legitimate relative filename that's 40+ characters of pure alphanumeric-plus-+/= with no dot, slash, backslash, or colon is vanishingly rare — filenames almost always have an extension or live under a directory separator. So this narrows false negatives to a genuinely contrived case (someone deliberately naming a file YWJjZGVmZ2hpams... with no extension) while it doesn't touch normal paths.

This sounds reasonable to me. The only way to get this to be 100% accurate is to separate the PEM string and file path into separate configuration options like the collector did but that's a breaking change.

transport/tlscommon/tls.go
AndersonQ: > A slightly malformed PEM file also turns out to be a potentially valid file path as well, unfortunately: https://go.dev/play/p/wgQ3QwzDZYM

Oh, that's interesting and unexpected

transport/tlscommon/tls.go
AndersonQ: > The only way to get this to be 100% accurate is to separate the PEM string and file path into separate configuration options like the collector did but that's a breaking change.

Yes, I was never a fan of this one-setting-does-all

transport/tlscommon/tls.go
AndersonQ: done

AndersonQ:
@cmacknz, when you have some time, please re-review

khushijain21 (approved):
Looks good to me!

andrzej-stencel (approved):

AndersonQ:
@cmacknz, when you have some time, please re-review

cmacknz (approved):

PR #433, opened 2026-07-10
Remove global logger from tlscommon package

khushijain21:
## What does this PR do?
This PR removes global logger instances from tlscommon package

## Why is it important?
To be able to run beatreceivers with their loggers colliding

### DIFF
--- a/transport/tlscommon/diag.go
+++ b/transport/tlscommon/diag.go
@@ -25,6 +25,8 @@ import (
 	"io"
 	"log"
 	"time"
+
+	"github.com/elastic/elastic-agent-libs/logp"
 )
 
 // DiagCerts returns a diagnostics hook callback that will validate if the certifiactes (cert + key, and CAs) present in the config are valid.
@@ -84,7 +86,7 @@ func diagCertificate(logger *log.Logger, cfg *CertificateConfig) {
 		logger.Print("certificate keypair is nil.")
 		return
 	}
-	crt, err := LoadCertificate(cfg)
+	crt, err := LoadCertificate(cfg, logp.NewNopLogger())
 	if err != nil {
 		logger.Printf("certificate keypair error: %v", err)
 		return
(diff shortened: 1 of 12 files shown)


transport/tlscommon/diag.go
```diff
@@ -84,7 +86,7 @@ func diagCertificate(logger *log.Logger, cfg *CertificateConfig) {
 		logger.Print("certificate keypair is nil.")
 		return
 	}
-	crt, err := LoadCertificate(cfg)
+	crt, err := LoadCertificate(cfg, logp.NewNopLogger())
```
orestisfl, line 92: Is this nop intentional?

khushijain21: Yes because, this method accepts a std logger which is incompatible with `*logp.Logger` type required by `LoadCertificate` . https://github.com/khushijain21/elastic-agent-libs/blob/remove-global-logger-tlscommon/transport/tlscommon/diag.go#L79-L107

And any errors from `LoadCertificate` will be logged here anyway. So we should good https://github.com/khushijain21/elastic-agent-libs/blob/5f472a8806765a62dab4bc7a948e7189b40ad7a4/transport/tlscommon/diag.go#L91

belimawr: I was also confused by it, but after @khushijain21's comment and with the helper of a clanker I can see it makes sense.

`diagCertificate` is used to generate a diagnostics for the certificates, when `LoadCertificate` is used to load certificates for application use, it receives a real logger.

@khushijain21 could you add a comment explaining why it is fine to use a no-op logger here?

khushijain21: Done. PTAL again when you can @orestisfl @belimawr

belimawr: I meant a comment explaining that `diagCertificate` is only used to generate diagnostics, any issues with certificates when the're being actually used are correctly logged when loading the certificates.

Anyways, that's not a big deal, so I'll approve the PR.

belimawr (approved):

macdewee:
@khushijain21 Seems like there are some lint issues introduced by this PR :crying_cat_face:

khushijain21:
> @khushijain21 Seems like there are some lint issues introduced by this PR 😿
 
Fixing them. On it

PR #435, opened 2026-07-13
Add LoggingWithTypedOutputsNonGlobal for non-global logger construction

samuelvl:
## What does this PR do?
Adds `LoggingWithTypedOutputsNonGlobal` and `ConfigureWithTypedOutputNonGlobal`, which create a typed-output logger and return it without replacing the global logger.

## Why is it important?
The existing `…Local` variants still mutate the global logger as a side effect, which breaks components that need their own independent logging configuration. These new functions make it safe to create component-specific loggers without interfering with the agent's main logging setup.

## Checklist

## Author's Checklist
- [ ]

## Related issues

-


samuelvl:
CI failures are unrelated to these changes, it was introduced in https://github.com/elastic/elastic-agent-libs/pull/433

logp/configure/logging.go
```diff
@@ -112,6 +113,40 @@ func LoggingWithTypedOutputsLocal(beatName string, cfg, typedCfg *config.C, logK
 	return logp.ConfigureWithTypedOutputLocal(config, typedLogpConfig, logKey, kind, outputs...)
 }
 
+// LoggingWithTypedOutputsNonGlobal is identical to LoggingWithTypedOutputsLocal but does not
+// mutate the global logger, making it safe for creating component-specific loggers.
+func LoggingWithTypedOutputsNonGlobal(beatName string, cfg, typedCfg *config.C, logKey, kind string, outputs ...zapcore.Core) (*logp.Logger, error) {
```
samuelvl, line 94: @khushijain21 @orestisfl what are your thoughts about this new constructor?

For context, in the agent we are going to need 2 different loggers (see [this PR](https://github.com/elastic/elastic-agent/pull/15491)). One for the agent and another for the otel collector (different paths for log files) so we cannot reuse the same global logger.

I think we should use `LoggingWithTypedOutputsLocal` but it's still global. I'm wondering if it is fine to introduce `LoggingWithTypedOutputsNonGlobal` and migrate to `LoggingWithTypedOutputsLocal`  when it is ready.

In the agent, despite using `LoggingWithTypedOutputsLocal` we are using the global logger to update the log level (see [code](https://github.com/elastic/elastic-agent/blob/62a52edb2378858c6eea21b20bc17df1c75aeb9f/pkg/core/logger/logger.go#L162 )). Do we have an epic/issue to track the migration? I can help with the migration to local loggers in the agent if needed.

khushijain21: @samuelvl this looks like a valid addition. 

There is an issue to track this migration for beats here https://github.com/elastic/ingest-dev/issues/6000 but not for elastic-agent

logp/configure/logging.go
```diff
@@ -112,6 +113,40 @@ func LoggingWithTypedOutputsLocal(beatName string, cfg, typedCfg *config.C, logK
...
+	config := logp.DefaultConfig(environment)
```
khushijain21: [nit] I think you can move the common part into into its own method

logp/core.go
```diff
@@ -340,6 +340,46 @@ func ConfigureWithTypedOutputLocal(defaultLoggerCfg, typedLoggerCfg Config, key,
...
+//   - `defaultLoggerCfg` is used to create a new core that will be the default
+//     output from the logger
+//   - `typedLoggerCfg` is used to create a new output that will only be used
+//     when the log entry matches `entry[logKey] = kind`
+//   - `key` is the key the typed logger will look at
+//   - `value` is the value compared against the `logKey` entry
+//   - `outputs` is a list of cores that will be added together with the core
+//     generated by `defaultLoggerCfg` as the default output for the loggger.
+func ConfigureWithTypedOutputNonGlobal(defaultLoggerCfg, typedLoggerCfg Config, key, value string, outputs ...zapcore.Core) (*Logger, error) {
+	sink, level, selectors, err := createSink(defaultLoggerCfg, outputs...)
```
khushijain21: same for this

logp/configure/logging_test.go
```diff
@@ -111,3 +115,139 @@ func TestLoggerOutputEnvironment(t *testing.T) {
...
+
+		logger.Debug("suppressed")
+		assert.Equal(t, 0, obsLogs.Len(), "Debug must be suppressed at InfoLevel")
+
+		logger.Info("visible")
+		assert.Equal(t, 1, obsLogs.Len(), "Info must reach the observer")
+	})
+}
+
+func TestLoggingWithTypedOutputsLocalWithLevel(t *testing.T) {
```
samuelvl: We need these `...WithLevel` variants so we can have a shared log level setter in the agent for base and otel logger. Both are controlled by the same `agent.logging.level` field.

logp/core.go
```diff
@@ -306,15 +309,70 @@ func ConfigureWithTypedOutput(defaultLoggerCfg, typedLoggerCfg Config, key, valu
...
+//   - `outputs` is a list of cores that will be added together with the core
+//     generated by `defaultLoggerCfg` as the default output for the loggger.
+func ConfigureWithTypedOutputNonGlobal(defaultLoggerCfg, typedLoggerCfg Config, key, value string, outputs ...zapcore.Core) (*Logger, error) {
+	level := zap.NewAtomicLevelAt(defaultLoggerCfg.Level.ZapLevel())
+	return ConfigureWithTypedOutputNonGlobalWithLevel(defaultLoggerCfg, typedLoggerCfg, key, value, level, outputs...)
+}
+
+// ConfigureWithTypedOutputNonGlobalWithLevel is like ConfigureWithTypedOutputNonGlobal but accepts a shared AtomicLevel
+// so the caller can change the log level at runtime without rebuilding the logger.
+func ConfigureWithTypedOutputNonGlobalWithLevel(defaultLoggerCfg, typedLoggerCfg Config, key, value string, level zap.AtomicLevel, outputs ...zapcore.Core) (*Logger, error) {
```
khushijain21: @samuelvl can you explain this? What do you mean "without rebuilding" the logger. If I follow the chain, it "is" building a new local logger.

Also, I don't feel very good about this design of "withLevel". Why can't we use `agent.logging.level` to control both base and otel logger separetely? I don't think I follow the reasoning for a shared logger level.

samuelvl: The reason was that the log level used by the baseLogger setter is global, leaving us with 2 options: 1) Have `Local`/`NonGlobal` loggers sharing the same global logger (which defeats their purpose) or 2) Define a shared log level that can be re-used by both loggers but requiring a way to pass it down, that's why I added the `...WithLevel` variants.

But this is not finally needed since we can apply changes to the collector logger dynamically: https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/manager/manager.go#L320

logp/configure/logging.go
samuelvl: done!

logp/core.go
samuelvl: done!

orestisfl (approved):

PR #436, opened 2026-07-15
feat(testing/fipsscan): add FIPS dependency scanner helper package

macdewee:
## Summary

Adds `testing/fipsscan` — a helper package for testing and documenting FIPS 140-3 compliance violations in Go binaries.

I would like to use it across different repos, so we can track the status of strict FIPS compliance adoption and prevent regressions

Go's certified FIPS 140-3 module (`GOFIPS140=v1.0.0`) only covers `crypto/*` stdlib packages. `golang.org/x/crypto` and other third-party crypto implementations are not covered. This package scans a binary's full dependency tree and surfaces every such import, attributing it to the responsible component and tracking known exceptions.

## Usage

```go
//go:build requirefips

package main_test

import (
    "testing"
    "github.com/elastic/elastic-agent-libs/testing/fipsscan"
)

var knownViolations = map[string]string{
    "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/sqlserverreceiver": "transitively imports x/crypto/md4 (Kerberos RC4-HMAC, NTLM)",
    "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/mongodbreceiver":  "transitively imports x/crypto/pbkdf2, x/crypto/scrypt, x/crypto/ocsp",
}

const (
    binaryPkg = "github.com/elastic/elastic-agent/internal/edot"
    rootPkg   = "github.com/elastic/elastic-agent/internal/edot"
)

func TestFIPSFullyCompliant(t *testing.T) {
    fipsscan.CheckViolations(t, binaryPkg, rootPkg, nil, knownViolations)
}
```

`CheckViolations` fails the test for any new unknown violation (must be added to `knownViolations` or the dependency removed) and for any stale entry in `knownViolations` that no longer reaches a forbidden package. Pass `extraForbiddenPkgs` to flag additional non-FIPS packages beyond `x/crypto`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,5 +21,6 @@ Provided packages:
 * `github.com/elastic/elastic-agent-libs/service` utilities to inspect services and collect debug data.
 * `github.com/elastic/elastic-agent-libs/str` the previous `stringset.go` file from `github.com/elastic/beats/v7/libbeat/common`. It provides a string set implementation. 
 * `github.com/elastic/elastic-agent-libs/testing` Testing helpers for network communication and outputs.
+* `github.com/elastic/elastic-agent-libs/testing/fipsscan` Test helper for auditing FIPS 140-3 compliance — scans a binary's dependency tree for forbidden (non-certified) crypto imports and maintains a known-violations allowlist. Requires the `requirefips` build tag.
 * `github.com/elastic/elastic-agent-libs/transport/tlscommon` TLS configuration and validation, CA pinning, etc.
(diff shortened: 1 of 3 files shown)


blakerouse (approved):
I was original going to say this seems like it doesn't belong in this repo, but then I went and checked the existing testing/ directory and see that other testing related cross-repo things are here. That makes sense then on why it is here!

PR #438, opened 2026-07-17
feat(testing/fipsscan): extend the fipsscan package to improve FIPS auto-audit

macdewee:
## TL;DR

Unifying the way we test binaries/modules for FIPS compliance. Aim was to just import this module, prepare list of known violations and Voilà. Example usages:

- https://github.com/elastic/fleet-server/pull/7395
- https://github.com/elastic/beats/pull/52036
- https://github.com/elastic/elastic-agent/pull/15600

vibe-coded. This PR is bigger than I expected and mostly resonate about resolving the dependencies between binary, component and library which violates the FIPS

Iterated few times to give satisfying results and minimal usage in fleet-server, elastic-agent and beats repo

## Summary

Extend `testing/fipsscan` package — a shared Go testing helper for auditing FIPS 140-3 compliance of Go modules. It scans dependency trees for imports of known non-FIPS crypto libraries and reports violations via the standard `testing.T` interface.

The package is guarded by `//go:build requirefips` so it compiles and incurs zero overhead outside FIPS-focused test runs (`go test -tags requirefips`).

## Motivation

FIPS compliance needs to be enforced across multiple Elastic repositories (elastic-agent, otel-collector-components, beats, …). Without a shared helper, each repo would reimplement violation detection independently and inconsistently. This package provides a single place to maintain the list of known non-FIPS libraries and a stable API for documenting accepted violations per component.

### `CheckModule` — the main entry point

```go
fipsscan.CheckModule(t,
    []string{"./..."},   // patterns passed to go list
    skipBinaries,        // non-shipped binaries to exclude
    nil,                 // extra forbidden prefixes beyond the baseline
    map[string]map[string][]fipsscan.KnownViolation{
        // outer key: binary import path or module-root prefix; "" = all binaries
        "github.com/elastic/elastic-agent": {
            // inner key: component — package path or module-root prefix in the chain
            "github.com/elastic/gokrb5/v8": {
                {Imported: "github.com/jcmturner/gofork", Reason: "Elastic gokrb5 fork depends on jcmturner gofork"},
                {Imported: "golang.org/x/crypto/md4",     Reason: "Kerberos RC4-HMAC requires MD4"},
            },
            "github.com/foxboron/go-tpm-keyfiles": {
                {Imported: "golang.org/x/crypto/chacha20poly1305", Reason: "TPM key parsing uses ChaCha20"},
            },
        },
    },
)
```

`CheckModule` reports undocumented violations via `t.Errorf` and flags stale entries (violations that no longer exist) so the map stays current.

### Key design points

- **Component-oriented config.** The two-level map (`binary → component → violations`) lets you document *which part of the binary* owns each violation. The component key is any package path or module-root prefix appearing in the import chain from the binary down to the direct importer — use a broad prefix like `"github.com/twmb/franz-go"` to cover all violations from that library in one entry, or an exact package for precision.
- **Prefix matching.** Both the outer binary key and `KnownViolation.Imported` support module-root prefixes. `{Imported: "github.com/jcmturner/gokrb5/v8"}` matches any `gokrb5/v8/…` sub-package. Binary key `"github.com/elastic/elastic-agent"` matches any binary under that module root.
- **Stale detection.** Entries whose violation no longer exists are flagged stale. Entries whose component was removed from the dependency tree entirely are silently skipped (not a stale error — the problem went away).
- **Library module support.** Falls back to flat violation detection (no binary attribution) when the scanned module has no `package main` entries.
- **`skipBinaries`.** Dev tools, test helpers, and non-shipped binaries can be excluded from the scan and from stale detection.
- **Single `go list` pass.** All patterns are resolved in one subprocess invocation regardless of how many binaries match.

### Lower-level functions

```go
// Per-binary scan with BFS attribution; Violation.Binary is set.
violations, importGraph := fipsscan.ScanBinaries(t, []string{"./cmd/..."}, skipBinaries, nil)

// Flat scan; Violation.Binary is empty.
violations, importGraph := fipsscan.Scan(t, "./...", nil)

// Display helpers.
chain := fipsscan.ShortestChain(v.Binary, v.Importer, importGraph)
fmt.Println(fipsscan.FormatChain(append(chain, v.Imported)))
```

### Baseline forbidden-package list

Ten third-party crypto libraries are checked by default (all exported as constants):

| Constant | Module prefix |
|---|---|
| `XCrypto` | `golang.org/x/crypto/` |
| `JcmturnerGokrb5` | `github.com/jcmturner/gokrb5/` |
| `JcmturnerAescts` | `github.com/jcmturner/aescts/` |
| `JcmturnerGofork` | `github.com/jcmturner/gofork/` |
| `XdgGoPbkdf2` | `github.com/xdg-go/pbkdf2` |
| `ProtonMailGoCrypto` | `github.com/ProtonMail/go-crypto/` |
| `CloudflareCircl` | `github.com/cloudflare/circl/` |
| `AzureGoNtlmssp` | `github.com/Azure/go-ntlmssp` |
| `YoumarkPkcs8` | `github.com/youmark/pkcs8` |
| `FilippioIO` | `filippo.io/` |

## Usage convention

The recommended filename is `fips_compliance_test.go` and function name `TestFIPSCompliance` in every adopting module. This makes ecosystem-wide auditing trivial:

```bash
grep -rl 'TestFIPSCompliance' .          # find all modules with a FIPS test
go test -tags requirefips -run TestFIPSCompliance ./...  # run them all
```

## Test plan

- [ ] `go build -tags requirefips ./testing/fipsscan/` passes
- [ ] `go test -tags requirefips ./testing/fipsscan/` passes (unit tests for `ShortestChain`, `FormatChain`, `matchesBinaryKey`, `isForbidden`, `moduleRoot`)
- [ ] Validate against real usage: `drosiek-fips` branch in elastic-agent uses this package for `elastic-agent` and `edot` binaries

### DIFF
--- a/testing/fipsscan/fipsscan.go
+++ b/testing/fipsscan/fipsscan.go
@@ -415,8 +415,14 @@ func FormatChain(chain []string) string {
 // Binaries in skipBinaries are excluded from the scan and from stale detection.
 func CheckModule(t testing.TB, patterns []string, skipBinaries []string, extraForbiddenPkgs []string, knownViolations map[string]map[string][]KnownViolation) {
 	t.Helper()
-
 	violations, importGraph, mains := scanBinaries(t, patterns, skipBinaries, extraForbiddenPkgs)
+	checkViolations(t, violations, importGraph, mains, knownViolations)
+}
+
+// checkViolations is the matching and stale-detection logic for CheckModule,
+// extracted to allow unit testing without a go list subprocess.
+func checkViolations(t testing.TB, violations []Violation, importGraph map[string][]string, mains []string, knownViolations map[string]map[string][]KnownViolation) {
+	t.Helper()
 
 	// matched[binKey][compKey][i] tracks whether knownViolations[binKey][compKey][i] was hit.
 	matched := make(map[string]map[string][]bool)
@@ -427,7 +433,7 @@ func CheckModule(t testing.TB, patterns []string, skipBinaries []string, extraFo
 		}
 	}
 
-	// Pre-compute reachable set per binary for stale detection.
+	// Pre-compute reachable set per binary for attribution and stale detection.
 	reachableFrom := make(map[string]map[string]bool, len(mains))
 	for _, bin := range mains {
 		reachableFrom[bin] = bfsReachable(bin, importGraph)
@@ -535,15 +541,21 @@ func CheckModule(t testing.TB, patterns []string, skipBinaries []string, extraFo
 		return keys
 	}
 
-	// markKnown finds ALL known entries where the component can reach importer
-	// and (Importer, Imported) matches. Marks each matched entry and returns the
-	// full list so the caller can log reasons. A single violation can match
-	// multiple component keys.
+	// markKnown finds ALL known entries where the component can reach the
+	// importer AND is reachable from the binary, and (Importer, Imported)
+	// matches. Marks each matched entry and returns the full list so the caller
+	// can log reasons. A single violation can match multiple component keys.
 	markKnown := func(binary, importer, imported string) ([]KnownViolation, bool) {
 		var results []KnownViolation
 		for _, binKey := range binKeysForBinary(binary) {
 			for compKey, kvs := range knownViolations[binKey] {
 				if !componentCanReach(compKey, importer) {
+					continue
+				}
+				// The component must also be reachable from the violation's binary.
+				// Without this check a component in a different binary can suppress
+				// violations here via the shared merged import graph.
+				if binary != "" && compKey != "" && !isReachableFromAny(compKey, []string{binary}, reachableFrom) {
 					continue
 				}
 				for i, kv := range kvs {
--- a/testing/fipsscan/fipsscan.go
+++ b/testing/fipsscan/fipsscan.go
@@ -17,8 +17,8 @@
 
 // Package fipsscan provides helpers for auditing import-policy compliance of Go
 // binaries by scanning their dependency trees for forbidden imports.
-// Callers supply the forbidden-package prefixes and known-violations allowlist;
-// the package itself is policy-agnostic and carries no build constraints.
+// The package is policy-agnostic: callers supply the forbidden-package prefixes
+// and the known-violations allowlist.
 package fipsscan
 
 import (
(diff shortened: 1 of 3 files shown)


testing/fipsscan/fipsscan.go
```diff
@@ -24,38 +24,142 @@
...
+	AzureGoNtlmssp = "github.com/Azure/go-ntlmssp"
+
+	// YoumarkPkcs8 covers github.com/youmark/pkcs8; handles PKCS#8 keys and
+	// may negotiate non-FIPS ciphers (RC2, 3DES, SM4) depending on the key file.
+	YoumarkPkcs8 = "github.com/youmark/pkcs8"
+
+	// FilippioIO covers filippo.io/ packages (edwards25519, age, mlkem768,
+	// etc.). None are part of a FIPS 140-3 certified module boundary, even
+	// when they implement a FIPS-standardized algorithm (e.g. FIPS 203 ML-KEM).
+	FilippioIO = "filippo.io/"
```
kruskall: This is excluding a whole namespace which looks too broad :(

kruskall (commented):
> FIPS compliance needs to be enforced across multiple Elastic repositories (elastic-agent, otel-collector-components, beats, …). Without a shared helper, each repo would reimplement violation detection independently and inconsistently.

I'm not convinced this is enough to justify this. Downstream components have different dependencies and they need to constantly audit their dependency graph (ideally even without fips). It would be more surprising if they all share the same forbidden packages.

elastic-agent-libs is a library, we shouldn't leak (more) implementation details of downstream consumers in here (despite how bad it already is). Trying to maintain a centralized list will add noise a maintenance burden to everyone.

testing/fipsscan/fipsscan.go
```diff
@@ -24,38 +24,142 @@
...
+		dir = parent
+	}
+}
 
-	cmd := exec.CommandContext(t.Context(), "go", "list", "-json", "-deps", "-tags", "requirefips", pkg)
+// goListPackages runs go list -json -deps -tags requirefips from the module root
+// and decodes the output. Shared by scanBinaries and Scan.
+func goListPackages(t testing.TB, patterns []string) []goListPackage {
+	t.Helper()
+	args := append([]string{"list", "-json", "-deps", "-tags", "requirefips"}, patterns...)
```
kruskall: consumers of this library use different tags so this needs to be configurable to work reliably

belimawr: +1

belimawr (commented):
So far I mostly vibe-reviewed it, my clanker found one issue:
```
  • High: CheckModule can suppress a violation for a component that is not reachable from the scanned binary. markKnown only checks that the component can reach the importer (componentCanReach), but never verifies that
    the component is reachable from v.Binary:
    • testing/fipsscan/fipsscan.go:462-468
    • testing/fipsscan/fipsscan.go:540-552

  Because component reachability is computed across the entire merged import graph, an unrelated component elsewhere in the module can document/suppress a violation from another binary whenever both eventually reach the
  same importer. This silently weakens the FIPS audit and contradicts the API documentation requiring both directions of reachability.

  markKnown should additionally verify that the component key matches a package reachable from the violation’s binary (except library-mode scans where Binary == ""). Add a regression test covering two binaries sharing an
  importer, with the known component reachable only from the other binary.

  The tagged package tests and PR CI pass, but they do not exercise this cross-binary attribution case.
```

Could you take a look at it? It seems legit. After that I'll do a more in-depth review.

macdewee:
@kruskall @belimawr I changed the approach to be fully stateless across projects and move the responsibility of providing the list of dependencies and tags on the caller's side

testing/fipsscan/fipsscan.go
```diff
@@ -64,39 +101,142 @@ func Scan(t testing.TB, pkg string, extraForbiddenPkgs []string) ([]Violation, m
...
+	for _, p := range pkgs {
+		if p.Name == "main" && !skip[p.ImportPath] {
+			mains = append(mains, p.ImportPath)
+		}
 		importGraph[p.ImportPath] = p.Imports
+	}
+
+	if len(mains) == 0 {
+		return findViolations(importGraph, forbiddenPkgs), importGraph, nil
+	}
```
belimawr, line 157: My clanker found an issue here:
```
Do not fall back to a library scan when every discovered binary is skipped

   testing/fipsscan/fipsscan.go lines 126-135

      for _, p := range pkgs {
          if p.Name == "main" && !skip[p.ImportPath] {
              mains = append(mains, p.ImportPath)
          }
          importGraph[p.ImportPath] = p.Imports
      }
      if len(mains) == 0 {
          return findViolations(importGraph, forbiddenPkgs), importGraph, nil
      }

  details:
  • binary mode: at least one discovered package main is not skipped. scanBinaries performs per-binary BFS and sets Violation.Binary.
  • library mode: len(mains) == 0. It calls findViolations over the complete merged graph and returns violations with an empty Violation.Binary.

  The ambiguity:
  • mains excludes skipped binaries while it is constructed.
  • Thus len(mains) == 0 means either:
    • the scanned patterns contain no package main packages: legitimate library mode; or
    • all discovered binaries were skipped: should produce no audit result, but is treated as library mode.

  CheckModule passes that result through without distinguishing the two cases:

   testing/fipsscan/fipsscan.go lines 342-346

  func CheckModule(t testing.TB, patterns []string, skipBinaries []string, forbiddenPkgs []string, tags []string, knownViolations map[string]map[string][]KnownViolation) {
      t.Helper()
      violations, importGraph, mains := scanBinaries(t, patterns, skipBinaries, forbiddenPkgs, tags)
      checkViolations(t, violations, importGraph, mains, knownViolations)
  }

  observable behavior:
  • CheckModule: all-skipped binaries generate library-mode violations (Violation.Binary == "") and may cause stale-entry errors.
  • ScanBinaries: all-skipped binaries instead fail with ScanBinaries: no package main found, because it interprets the same empty mains value as no main was found.

   testing/fipsscan/fipsscan.go lines 204-207

      violations, importGraph, mains := scanBinaries(t, patterns, skipBinaries, forbiddenPkgs, tags)
      if len(mains) == 0 {
          t.Fatalf("ScanBinaries: no package main found matching %v", patterns)
      }

  fix:
  • Track discoveredMains before applying skipBinaries.
  • Select library mode only when len(discoveredMains) == 0.
  • When binaries were discovered but all were skipped, return no violations and mark the scan as binary-mode for stale-detection purposes.

```

testing/fipsscan/fipsscan.go
```diff
@@ -160,44 +300,308 @@ func FormatChain(chain []string) string {
...
+
+	compReach := newReachabilityCache(importGraph)
+
+	// componentCanReach reports whether the component identified by key can
+	// reach importer via forward edges in the import graph. key "" is a wildcard.
+	componentCanReach := func(key, importer string) bool {
+		if key == "" {
+			return true
+		}
+		for _, pkg := range compPkgs[key] {
```
belimawr, line 426: Another finding from my clanker that seems legit:

```
Bound reachability-cache memory for broad component prefixes — testing/fipsscan/fipsscan.go:396

componentCanReach calls compReach.from for every package matched by a component prefix. reachabilityCache.from retains a complete reachable-set map for each such package for the remainder of the test. A module-root component key can match hundreds or thousands of packages; when its importer is unreachable, this retains up to one graph-sized map per package (O(component packages × graph packages) memory) and repeats the traversals. This can make the compliance test disproportionately memory-intensive on large consumer modules. Reverse reachability from each importer, intersected with the component and binary reachable sets, avoids retaining per-package full-graph maps.
```

belimawr (dismissed):
LGTMC - Looks Good to My Clanker!

Let me know if you want a detailed human-review on this one. It's not my area of expertise, but I'll do my best.

macdewee:
@belimawr I started doubting if this is not overengineered already (typical vibe-coded), will give it another local round of simplification and will see. Will update EOD

macdewee:
@belimawr Could you re-review please?

testing/fipsscan/fipsscan.go
```diff
@@ -65,85 +147,320 @@ func Scan(t testing.TB, pkg string, extraForbiddenPkgs []string) ([]Violation, m
...
-		pkg  string
-		path []string
+	for _, bin := range binaries {
+		for pkg := range bfsReachable(bin, graph) {
+			unionReach[pkg] = true
+			for _, imp := range graph[pkg] {
+				if !isForbidden(imp, forbiddenPkgs) {
+					continue
+				}
+				v := Violation{Binary: bin, Imported: imp}
```
belimawr, line 371: This does not include the `Importer`, which can cause a problem later on.
If we have:
```
cmd/agent ──→ component-a ──→ forbidden/pkg
cmd/agent ──→ component-b ──→ forbidden/pkg
```

And we want to allow only `cmd/agent ──→ component-b ──→ forbidden/pkg`, the test will end up passing even though we have the forbidden import `cmd/agent ──→ component-a ──→ forbidden/pkg` because the importer is omitted thus comparing only `Binary` and `Imported`.

It seems to me we should be able to make this distinction.

